PR #1, opened 2015-02-01
Add testing support

dgiagio:
- Improved .gitignore entries for IntelliJ IDEA and testing output
- Added/improved ant's support for testing and report generating
- Refactored some classes to allow for pluggable HTTP connection implementation (HttpURLConnection and HttpComponents [TODO])
- Added initial testing via JUnit
- Added support for mocking JSON responses on unit-tests
- Added support for mocking exceptions on unit-tests


renner:
Thanks for your pull request! I will review and give it a try to provide you some feedback later this day.

build.xml
```diff
@@ -11,6 +11,7 @@
   <property name="dir-lib" value="lib" />
   <property name="dir-bin" value="bin" />
   <property name="dir-dist" value="dist" />
+  <property name="dir-test-report" value="test-output" />
```
renner: Please follow the pattern here: `name="dir-*" value="*"`, e.g. change then name to `name="dir-test-output"`.

renner:
Ok, here is some feedback after an initial review of your patch:

I generally like the efforts to allow for pluggable HTTP connection implementation. Why don't you use this PR to continue on that path to actually introduce an Apache HttpComponents based implementation of the `SaltStackConnection`?

About the testing part I think we need something more than what you are suggesting, since your `SaltStackMockConnection` works around all the HTTP stack completely. Your test actually covers only the parsing of JSON. I guess we rather want to verify here that we are actually sending the right HTTP requests and then mock responses accordingly using a library or framework e.g. the one that is suggested by @skhtmp in https://github.com/SUSE/saltstack-netapi-client-java/pull/2.

I think I might give the library that is proposed in #2 a try, while for this PR I would suggest that you focus only on using HttpComponents for making the connections. Is that fine with you? Such kind of tests would also not care about what HTTP library we use (should work with both implementations equally), so testing efforts should actually not interfere very much with this PR.

Apart from that here are some general remarks:

- Make sure that every commit message describes what is actually done while avoiding repeating commit messages (like "Added/improved testing support").
- Smaller patches are easier to review and therefore can also be merged faster (for instance I might be fine with your changes to `.gitignore`, but I won't merge since we are still discussing the rest of the patch), so consider to make separate smaller PRs for unrelated patches.
- Respect a maximum line length of 92 characters (break longer lines and indent with 8 spaces).
- Make sure that at least all public methods have Javadoc comments in place.

Further please find and consider my inline comments as well.

src/main/java/com/suse/saltstack/netapi/client/SaltStackJdkConnection.java
```diff
@@ -0,0 +1,110 @@
...
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.lang.reflect.Type;
+import java.net.HttpURLConnection;
+
+/**
+ * Class representation of a connection to SaltStack for issuing API requests using JDK's HttpURLConnection.
+ */
+public class SaltStackJdkConnection implements SaltStackConnection {
```
renner: I know that it doesn't really matter, but for 3 letter acronyms (that cannot be [avoided] (http://www.oracle.com/technetwork/java/codeconventions-135099.html)) I would prefer to put them all uppercase in class names. So consider to rename this class (and related ones accordingly) to `SaltStackJDKConnection` or maybe get rid of JDK in the class names overall.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -74,11 +89,24 @@ public void setProxy(SaltStackProxySettings settings) {
      */
     public SaltStackToken login(String username, String password)
             throws SaltStackException {
+        return login(username, password, "auto");
```
isbm: I would rather suggest to avoid strings and document them in this case, as this param is actually SaltStack method to connect, perhaps?

```java
// Documentation, describing what does that mean
public class SaltStackAPIConnectionConstants {
    public static final AUTO = "auto";
    public static final OTHER_POSSIBLE_CONNECTION = "other_possible_connection";
}

...

    return login(username, password, SaltStackAPIConnectionConstants.AUTO);
```

src/test/java/com/suse/saltstack/netapi/client/SaltStackMockConnection.java
```diff
@@ -0,0 +1,35 @@
...
+    public SaltStackMockConnection(String mockJson) {
+        this.mockJson = mockJson;
+        this.mockException = null;
+    }
+
+    public SaltStackMockConnection(SaltStackException mockException) {
+        this.mockJson = null;
+        this.mockException = mockException;
+    }
+
```
isbm: I am puzzled here why we would need these _two_ constructors _either_ way? What would be wrong with only one constructor (in case we would need the entire class at all, which I am not sure at the moment)?
```java
public SaltStackMockConnection(String mockJson, SaltStackException mockException) {
   ...
}

SaltStackMockConnection a = new SaltStackMockConnection(null, ....);
SaltStackMockConnection b = new SaltStackMockConnection(..., null);

File: src/test/java/com/suse/saltstack/netapi/client/SaltStackMockConnection.java [hunk 5]
```diff
@@ -0,0 +1,35 @@
...
+    public <T> T getResult(Type resultType, String data) throws SaltStackException {
+        if (mockJson != null) {
+            return new Gson().fromJson(mockJson, resultType);
+        }
+        if (mockException != null) {
+            throw mockException;
+        }
+        throw new RuntimeException("Mock not configured properly");
+    }
+}
```
isbm: As @renner already pointed out, we would need here more than just parsing JSON.

src/main/java/com/suse/saltstack/netapi/client/SaltStackJdkConnection.java
```diff
@@ -0,0 +1,110 @@
...
+                outputStream.close();
+            }
+
+            // React depending on the result of the request
+            int responseCode = connection.getResponseCode();
+            if (responseCode == HttpURLConnection.HTTP_OK ||
+                    responseCode == HttpURLConnection.HTTP_ACCEPTED) {
+                inputStream = connection.getInputStream();
+                Reader inputStreamReader = new InputStreamReader(inputStream);
+                Reader streamReader = new BufferedReader(inputStreamReader);
```
isbm: Input streams has to be always closed, especially in `HttpURLConnection` because there are lots of things happening internally in there, such as keep-alive, connection buffering options etc.

build.xml
dgiagio-druid: OK, will do.

src/main/java/com/suse/saltstack/netapi/client/SaltStackJdkConnection.java
dgiagio-druid: I will follow your suggestion and focus on HttpComponents implementation, so this class will probably be removed.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
dgiagio-druid: Makes sense.

src/test/java/com/suse/saltstack/netapi/client/SaltStackMockConnection.java
dgiagio-druid: It was a matter of choice. There's nothing special with two constructors over just one.

src/test/java/com/suse/saltstack/netapi/client/SaltStackMockConnection.java
dgiagio-druid: Yes, makes sense.

src/main/java/com/suse/saltstack/netapi/client/SaltStackJdkConnection.java
dgiagio-druid: Even though this class has a new name, this is already existing code. Anyway, I'll be focusing my attention to HttpComponents implementation as suggested by @renner.

renner: @isbm: And stream is actually closed, see the implementation of `closeQuietly()` that's called in line 107.

isbm: Ah, right, I overlooked that. Not my code. :-)

PR #2, opened 2015-02-01
Junit restdriver tests

skhtmp:
* JUnit tests against mock server with rest-driver-client
* Implement POST /run
* command line integration tests


skhtmp:
For running JUnit 4 Test with ant-junit see also https://bugzilla.opensuse.org/show_bug.cgi?id=915686

renner:
Thanks for your pull request! I will give you some feedback later today after reviewing the code and trying it out.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -72,12 +72,12 @@ public void setProxy(SaltStackProxySettings settings) {
...
      * @throws SaltStackException if anything goes wrong
      */
-    public SaltStackToken login(String username, String password)
+    public SaltStackToken login(String username, String password, String eauth)
             throws SaltStackException {
         JsonObject json = new JsonObject();
         json.addProperty("username", username);
         json.addProperty("password", password);
-        json.addProperty("eauth", "auto");
+        json.addProperty("eauth"   , eauth);
```
renner: Why there is spaces before the comma here? Please remove.

src/test/java/com/suse/netapi/client/SaltStackClientMockTests.java
```diff
@@ -0,0 +1,111 @@
...
+import com.github.restdriver.clientdriver.ClientDriverRule;
+import com.github.restdriver.clientdriver.ClientDriverRequest.Method;
+import com.github.restdriver.clientdriver.ClientDriverRequest;
+
+import static com.github.restdriver.clientdriver.RestClientDriver.giveResponse;
+import static com.github.restdriver.clientdriver.RestClientDriver.onRequestTo;
+
+import com.github.restdriver.clientdriver.exception.ClientDriverFailedExpectationException;
+
+import com.suse.saltstack.netapi.exception.SaltStackException;
```
renner: In order to compile this in my IDE I had to add 3 more import statements:

    import com.suse.saltstack.netapi.client.SaltStackClient;
    import com.suse.saltstack.netapi.client.SaltStackRunResults;
    import com.suse.saltstack.netapi.client.SaltStackToken;

src/test/java/com/suse/netapi/client/SaltStackClientMockTests.java
```diff
@@ -0,0 +1,111 @@
+package com.suse.saltstack.netapi.client;
```
renner: The package is not matching the folder structure, please fix the folder structure in this case!

src/test/java/com/suse/netapi/tests/SaltStackClientIntegrationTests.java
```diff
@@ -0,0 +1,37 @@
+package com.suse.saltstack.netapi.tests;
```
renner: The package is not matching the folder structure as before, please fix the folder structure (package is fine).

renner:
Ok, here is some feedback after a first review of your proposed changes:

I am willing to try out that library that you are suggesting for the tests, even though I don't really like the fact that it brings in so many (testing) dependencies. Can we maybe omit some of the jars since we only want to use the client part of that library or do we need to have them all around?

Anyways, I have some problems testing your changes:

- The "testreports" folder apparently needs to be created manually for the first time (should be created in "init-tests" target instead and removed with "clean").
- I get a `Building MANIFEST-only jar` warning for the "dist-tests" target and the test actually fails for me. Please make sure that `ant` on the commandline *just works*. (BTW: thanks for the link to the bug report, I was suffering from it this morning!)
- Please fix all the errors that were pointed out in my previous line comments, I wonder how it can work for you. Maybe you forgot to push some commits?

Apart from that here are some general (but important) remarks. Please fix your patch accordingly:

- Your indentation settings seem to be messed up: Please **never** use tabs, indentation should be 4 spaces!
- Lines should not be longer than 92 characters, feel free to break lines and indent 8 spaces after a line break.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
skhtmp: It lined up nicely with the other commas ;-) Fixed.

src/test/java/com/suse/netapi/client/SaltStackClientMockTests.java
skhtmp: Did the IDE fix the folder-structure/package mismatch by adjusting the package statement and not the folder structure?

My intention was to have the test class in the same package as the class being tested, so these imports weren't necessary.

src/test/java/com/suse/netapi/client/SaltStackClientMockTests.java
skhtmp: Fixed.

skhtmp:
Apologies for the sloppiness, and thanks for taking the time to review this PR anyway. I hope I have caught all problems now.

Ant build: I don't know what the exact problem was, but changes in the tests only made it into the actual jar after running `dist-tests` _twice_. Most of the time. I was obviously confused when to include or exclude `main` and `test` subdirectories, so I refactored `build.xml` to more clearly handle main and test artifacts. I hope it is easier to understand now what's going on.

I did _not_, however follow this pattern and create `lib/main` and `lib/test` instead of `lib` and `testlib` -- I didn't want to break the convention of having one top-level `lib` directory for all run-time dependencies. 

I have considered adding a `dist-sources-no-tests` target as well, without test sources and dependencies, but this would have meant having two build files, or putting a `build.xml` with non-working targets into the source tarball, so I stopped there -- this is just a rough proposal after all.

Dependencies: I have stripped them down as much as possible. Another alternative might be MockServer, which has a similar API and is (optionally) distributed in one jar file containing all dependencies -- this would then be only one file to track, but it is about twice as large as the libraries I use now. 

REST-driver is only distributed in separate jar files through Maven central (and source of course), and dependencies are meant to be handled by Maven.

Also, I haven't looked at all possible HTTP server mocking solutions out there. There might well be something better and more light-weight which I missed. If this is the direction you'd like to take, I could do more research...

(From a textbook perspective I wouldn't even call these tests unit tests, as they test the full implementation from method call to HTTP request, response and back. It was the most interesting aspect to test, though ;-), so I started with that, and being able to automate it with JUnit is helpful whatever the textbooks say.)

renner:
Thanks for the fixes, it looks good! I would want to merge some of your changes, but I'm not yet decided what is the best library to use for the tests. May I ask you to make several smaller pull requests (rebased to the current master), so we could merge the other changes right away?

Separate pull requests could be:

1. The changes to build.xml
2. The implementation of `run()`
3. Tests using your preferred library

Thank you in advance!

skhtmp:
Will do, thanks for the feedback!

renner:
What I think should be tested is that we are sending out the right requests and that we get results (objects) as expected for given JSON examples according to the salt API documentation. If this is  combined we have rather an integrational roundtrip than a traditional unit test.

BTW: In the past we used http://www.simpleframework.org/ and we really liked it, see https://github.com/susestudio/susestudio-lib-java/pull/10/files.

PR #4, opened 2015-02-03
Improved .gitignore for IntelliJ IDEA

dgiagio:
Ignore IntelliJ IDEA workspace files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,3 @@
 # Build and tests
 /bin/
 /dist/
-/test-report/
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,13 @@
+# Eclipse
 /.classpath
 /.project
 /.settings/
+
+# IntelliJ IDEA
+.idea
+*.iml
+
+# Build and tests
 /bin/
 /dist/
+/test-report/


.gitignore
```diff
@@ -1,5 +1,13 @@
...
 /.settings/
+
+# IntelliJ IDEA
+.idea
+*.iml
+
+# Build and tests
 /bin/
 /dist/
+/test-report/
```
renner: There is no test reports produced so far, I would be fine with merging the IntelliJ part though. The test results should be added to .gitignore with a PR that actually generates such results, so please remove it from this patch and add it to another PR accordingly.

renner:
Note that every pull request should have a description. Please add a description to all of your PRs!

dgiagio:
@renner Done, thanks for reviewing.

renner:
Manually merged into master. I adjusted your new patterns a little in order to make them match more precisely, see 5aec108540c852d2c728437024b1e7a145f94e5a. I hope it still works for you, otherwise let me know (couldn't test it since I don't use IDEA).

dgiagio:
It's good, thanks!

PR #5, opened 2015-02-03
Added basic Ant + JUnit testing support

dgiagio:


renner:
Please add a description here as well. I can of course guess what you want to do, but please describe what is the purpose of this PR.

Apart from that it would be really nice if we could have at least a small unit test with this PR doing something simple, so we can see that the "testing support" actually works. This could be e.g. a test class for the methods in SaltStackClientUtils or similar. Adding test results to `.gitignore` should also go here.

build.xml
```diff
@@ -78,24 +80,23 @@
...
   <!-- Run tests using built artifacts -->
-  <target name="run-tests" depends="dist-jar, dist-tests">
+  <target name="run-tests" depends="dist-tests">
     <junit printsummary="yes" haltonfailure="yes">
+      <formatter type="plain" />
       <classpath refid="classpath-tests" />
-      <!-- Add tests here -->
+      <batchtest fork="yes" todir="${dir-test-report}">
+        <fileset dir="${dir-src-test}" includes="**/*Test.java"/>
+      </batchtest>
```
renner: I like it slightly better to list the single test classes here (using the `test` target), at least as long as there are only a few of them:

    <test name="com.suse.saltstack.netapi.client.FoobarTest" todir="${dir-test-results}">
        <formatter type="plain" />
    </test>

dgiagio:
@renner Done, thanks for reviewing.

dgiagio:
@renner JUnit is missing hamcrest. Below is the result of ```ant clean run-tests```

```
Testcase: initializationError took 0.002 sec
	Caused an ERROR
org/hamcrest/SelfDescribing
java.lang.NoClassDefFoundError: org/hamcrest/SelfDescribing
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:760)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:408)
Caused by: java.lang.ClassNotFoundException: org.hamcrest.SelfDescribing
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
```

renner:
It's not needed by our code and works fine on the commandline, I can reproduce your error only when I try to run it in Eclipse. It didn't happen for me though: I was using the JUnit 4 library that's included with Eclipse and it contains hamcrest as well. I reverted my commit with 685432ac19fe07e7d9aa671c381024123085517e .

PR #6, opened 2015-02-03
Add pluggable HTTP connection implementations support

dgiagio:
- Add initial infrastructure for multiple HTTP connection implementations.


dgiagio:
Next PR will have support for HttpComponents.

renner:
Please provide a description for this PR as well. Since your "next PR" will apparently be very dependant on this one (e.g. it can only be tested after this one is actually merged, changes in the next one might require API changes in this one, and so on) consider to keep these changesets together.

dgiagio:
- Added support for Apache's HttpClient, including Proxy support (authenticated or not)
- Tested Proxy support on my local Squid. Not sure how to mock a Proxy server yet.

moio:
@dgiagio, I cannot make sense of commit messages in this PR. Two have exactly the same comment, "Add support for Apache HttpClient as HTTP connection backend", despite one being mostly about code and the other mostly about tests, then there is "Fix small conflict on build.xml" which touches gitignore and java files, but not `build.xml`.

Can you please rebase/reword them in such a way they are consistent with their contents? Thanks!

moio:
@dgiagio further, I am confused about the relationship between this PR an PR #7. It looks like PR #7 introduces WireMock jars, and this one requires them, but I don't see PR #7's commits in this PR.

Maybe you want to organize it the other way round - namely adding code in this PR and tests in PR #7 - just please make sure that each one compiles/runs correctly separately.

dgiagio:
@moio Makes sense. I'll do, thanks.

PR #7, opened 2015-02-03
Add initial support for RESTful unit-tests via WireMock

dgiagio:
- Add support for RESTful testing via WireMock.
- Added /login test.


dgiagio:
@isbm done, thanks.

PR #9, opened 2015-02-04
Add initial support for RESTful unit-tests via WireMock

dgiagio:
- Add WireMock jar and slf4j minimal dependency
- Add unit-test for /login using WireMock.


renner:
Would you mind to give us a bit of reasoning about why you think we should use especially the WireMock library for testing? We want to know your opinion, so maybe you could explain your decision by answering the following questions:

- Why is WireMock better than other available libraries?
- When comparing it to e.g. [REST-driver] (https://github.com/rest-driver/rest-driver/wiki) (that has been proposed as well), what are the advantages or disadvantages?
- Is there any other reasons why you think we should use WireMock?

dgiagio:
I chose WireMock mostly because:
- It's one of the most popular libraries for HTTP mocking, it's actively developed and well-documented.
- WireMock is more popular than REST-driver (370+ stars vs 90+ stars on GitHub), has many more features (stub delays for timeout simulation, stateful behavior, can be used to record real service responses to build stubs, and so on), much better documentation and better API design in my opinion.
- My main goal was to pick a mock library that has enough features and good documentation. I think WireMock fits the bill.

On other hand, I have no problem using other library if you guys decide so.

renner:
Thanks a lot for your reasoning! I like the library and this PR, even though I might prefer to verify our requests separately from the responses which can be done like [this] (http://wiremock.org/verifying.html) it seems. Please find some additional inline comments as well.

Note that there is no hurry, I will wait with merging this to see the proposal by @skhtmp as well.

src/test/java/com/suse/saltstack/netapi/utils/SaltStackClientTest.java
```diff
@@ -0,0 +1,65 @@
+package com.suse.saltstack.netapi.client;
```
renner: This package does not match the file system path above, please fix the path in this case.

dgiagio:
@renner You're correct, the file system path was incorrect. Fixed, thanks!

renner:
We eventually decided to use WireMock, so I merged this. Thanks for your contribution!

dgiagio-druid:
Thanks!

PR #10, opened 2015-02-05
Refactor build.xml

skhtmp:
Clearer separation of "main" and "test" artifacts, hopefully
better readability.

(As discussed, sorry for the delay.)


renner:
Thanks for the contribution, I merged it. This makes the creation of jar files more straight forward IMO since no filtering or intermediate cleaning of the bin dir is needed. Further it also separates the ant generated .class files from those that are generated by an IDE (as long as "clean" is not called).

Note that there were some trailing spaces that I removed with e0f7ec002dee25dda1a02514d9eab2d2c2cecebd.

PR #8, opened 2015-02-04
Add support for multiple HTTP connection backends and Apache's HttpClient backend support 

dgiagio:
- Add initial infrastructure for multiple HTTP connection backends.
- Add support for Apache HttpClient as HTTP connection backend.
- Apache HttpClient backend supports proxy, authenticated or not.

PR #11, opened 2015-02-08
Added support for /keys endpoint

Oranor:
Hi,

I did some hacking and changed the following:

- made the eauth scheme configurable, so that other schemes than "auto" can be used for authentication.
- pulled the parsing of api call results out of the connection class to allow parsing of more complex results.
- implemented the /keys endpoint as an example.
- added some unit tests.


Oranor:
Oops you are using Java 7. The following snippet from SaltStackKeyParser works only with Java 8:

            if ( !keyResult.getClass().getTypeName()
                .equals(resultType.getTypeName())) {
            throw new SaltStackParsingException("Only SaltStackKeyResult can be parsed with this class");
        }
To get it running on Java 7 you can replace it with:

            if ( !"class com.suse.saltstack.netapi.client.SaltStackKeyResult"
                .equals(resultType.toString())) {
            throw new SaltStackParsingException("Only SaltStackKeyResult can be parsed with this class");
        }

src/main/java/com/suse/saltstack/netapi/parser/SaltStackKeyParser.java
```diff
@@ -0,0 +1,63 @@
...
+        Reader inputStreamReader = new InputStreamReader(inputStream);
+        Reader streamReader = new BufferedReader(inputStreamReader);
+
+        JsonParser parser = new JsonParser();
+        try {
+            JsonElement resultElement = parser.parse(streamReader);
+            JsonObject resultObject = resultElement.getAsJsonObject();
+            JsonObject returnObject = resultObject.getAsJsonObject("return");
+            for (Map.Entry<String, JsonElement> elementEntry : returnObject.entrySet()) {
+                if ("local".equals(elementEntry.getKey())) {
```
isbm: http://en.wikipedia.org/wiki/Yoda_conditions

Oranor: I didn't know this coding style had such a funny name. Thanks for the link. It became a habit for me since I am on a team that favors it.

renner:
Hello and thanks for your pull request! This request is quite big though and contains many changes that are unrelated to each other. I would therefore ask you to split up your request into separate smaller ones that can more easily be reviewed and merged. Also make sure to always rebase your branches to the current master in order to avoid conflicts when we want to test the proposed changes or merge them.

Further please also note the following points:

- The eauth scheme has already been made configurable with merging #8, so this part of your patch is probably obsolete.
- About the testing: I don't think we want the unit tests to require an actual salt master with API to be around. Rather we just want to verify that the library is sending the *correct* requests and that it can handle expected responses correctly. We might want to use a framework for that as proposed in #9 and #12. Feel free to make another proposal about that, but considering the fact that there is already two requests trying to introduce such tests I would probably focus on proposing something else first.
- Separating the code to parse the results from the connection code is IMO a very good idea. I would suggest you to go ahead with this and make a new PR containing those changes only.
- Implementing support for the */keys* endpoint is a good idea as well. I wouldn't consider this an example to test the code for parsing results though, rather this is a new feature for the library and should be done per separate pull request.

Feel free to contact me in case you have any questions.

Thank you and regards,
Johannes

src/main/java/com/suse/saltstack/netapi/parser/SaltStackKeyParser.java
```diff
@@ -0,0 +1,66 @@
...
+
+    private SaltStackKeyResult keyResult = new SaltStackKeyResult();
+
+
+    @Override
+    public <T> T parse(Type resultType, InputStream inputStream) throws SaltStackParsingException {
+        // uncomment this on Java 8
+        //if ( !keyResult.getClass().getTypeName()
+        //        .equals(resultType.getTypeName())) {
+        if ( !"class com.suse.saltstack.netapi.client.SaltStackKeyResult"
```
moio: Some other points:

* in general, please make sure your code compiles and works as intended in Java 7 for this project, as we are not sure if we will be able to support Java 8 yet. Sorry not to have mentioned that upfront, hopefully it's not too much work to adapt code to the older language;
* second, can you propose some alternative design in which reflection is not used to check types/raise exceptions at runtime? I guess for example that having a parametric `ISaltStackResultParser` interface would make more sense than a parametric `parse` method, as each concrete implementation would only handle a specific type.

Thanks!

Oranor:
Thanks for the thorough reviews. I will incorporate your suggestions and submit two new pull requests. One with a separated results parser, and one with the /keys endpoint.

PR #12, opened 2015-02-09
Junit mockserver

skhtmp:
As dependencies are an issue, I've looked at MockServer which is also available in one standalone jar. To be honest, it is very similar to WireMock, so I personally would be fine with using either of them. WireMock is a bit smaller, too.

I've also split the login test case in two, testing request sent and response handled separately.

### DIFF
--- a/src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
+++ b/src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
@@ -76,10 +76,8 @@ public class SaltStackClientTest {
             .respond(
                 response()
                     .withStatusCode(200)
-                        .withHeaders(
-                            new Header("Content-Type", "application/json")
-                        )
-                        .withBody(JSON_LOGIN_RESPONSE)
+                    .withHeaders(new Header("Content-Type", "application/json"))
+                    .withBody(JSON_LOGIN_RESPONSE)
             );
 
         try {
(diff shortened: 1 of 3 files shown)


src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
```diff
@@ -0,0 +1,96 @@
...
+            )
+            .respond(
+                response()
+                    .withStatusCode(200)
+                        .withHeaders(
+                            new Header("Content-Type", "application/json")
+                        )
+                        .withBody(JSON_LOGIN_RESPONSE)
+            );
+
```
isbm: Very interesting indentation. :-)

skhtmp: Indeed.

renner:
Hello and thanks for your contribution! Indeed both of the proposed libraries seem to be very similar to each other, at least feature wise. We decided now to use wiremock though, and honestly: this is mainly because wiremock seems to be the more popular option of both.

I want to point out that our decision has nothing at all to do with the coding itself, your coding is truly fine! Rather it is really only about choosing a library. I further like how you are suggesting to test the requests separately from the responses, I might apply this or a similar pattern to the tests later on.

Note also that dependencies are not much of an issue either, *standalone jars* usually simply self-contain their dependencies, which is technically the same as including them separately.

Please feel encouraged to come up with other pull requests, your efforts are very much appreciated. Thanks again!

skhtmp:
Thanks for the update ;-)

I meant: handling dependencies in separate jars seems to be an issue, and a standalone jar is more comfortable.

I'd _actually_ prefer having all test libraries separate from the runtime libraries, and separate jar files so that it is clear what is included and which version of which library is used where -- but I can also understand your point.

PR #13, opened 2015-02-09
Make resultType stricter to catch errors at compile time

lucidd:
This makes writing 
```java 
SaltStackTokenResult result = connectionFactory.create("/login", config).getResult(Integer.class, json.toString())
```
a compile time error instead of a runtime error.


moio:
LGTM, a step in the right direction!

Thanks for your contribution.

PR #14, opened 2015-02-11
Pulling the results parsing out of the connection classes

Oranor:
I pulled the result parsing out of the connection classes to allow customized parsers for complex results from the netapi. This is an updated version of a previous pull request.

I liked the idea of a parametric Interface that was suggested after my last pull request. But I changed it to an abstract base class. This allows reusing common methods by multiple parsers.


moio:
Thanks for this further PR!

I am still a bit unsure about the design of these classes.

Is there any way to avoid passing the `resultType` around, maybe using `T` instead?

Thanks!

Oranor:
I changed it now. My result parsers don't need resultType anymore, so it can get removed from the getResults() call in SaltStackConnection.

moio:
I wonder if it can be removed from Parser classes themselves. Is that possible in your opinion?

Oranor:
It would be possible if we pull down the parseSimple method into the derived classes. But then we would have to maintain the same code at three places, which would violate the DRY principle.

Alternatively we could use JsonParser, like I did in my first submission. But that would need more complicated code that need to be done differently for each of the result types. 

I don't see an advantage in any of these alternatives.

lucidd:
@moio  Sadly we can't use `T` itself to get the Type at runtime because of how Javas generics currently work (with the exception of a little "hack" using anonymous classes but that comes with its own set of problems). The only thing we can do is push the problem of passing in the right type to a point where we don't have to do it every time we use the parser (just like @Oranor did with hard coding the type information in each subclass of the parser).

I have a simplifications to propose to this PR and a few changes that build on top of it. The only thing that's different for each parser is the type it needs to parse. The parsing can be implemented in the `SaltStackParser` itself while all Subclasses just pass a different Type to the super Constructor. 

```java
public class SaltStackParser<T> {

    private final Class<T> type;

    public SaltStackParser(Class<T> type){
        this.type = type;
    }

    public T parse(InputStream inputStream) throws SaltStackParsingException {
        Reader inputStreamReader = new InputStreamReader(inputStream);
        Reader streamReader = new BufferedReader(inputStreamReader);

        // Parse result type from the returned JSON
        return new Gson().fromJson(streamReader, type);
    }
}
```

This also allows us to create parsers without the need to create a subclass for it, which at that point would be more boilerplate then "logic" anyway.

Building on top of this proposal i experimented with two additional changes which give the result types the same treatment and reducing them to one. There are also some changes in the Connection  classes to make use of the type parameters and make sure they line up correctly.

All the changes i made can be found [here](https://github.com/lucidd/saltstack-netapi-client-java/compare/Oranor:resultparser...wip) but i will hold of on submitting a PR until this PR is resolved since mine is based on it.

moio:
@Oranor: I merged your pull request with a couple of small nitpicks. Thanks a lot for your contribution!

@lucidd: your answer is very correct.

For us to review your further proposals it is better to have a pull request, so now that @Oranor's code has been accepted you are encouraged to open one.

PR #15, opened 2015-02-12
Use descriptive method name, not API endpoint

skhtmp:
At the moment, the methods in SaltStackClient are called like the salt netapi endpoints they are connecting to. If this method naming scheme is followed, at some point we well have overloaded methods of the same names calling their netapi endpoints with POST and GET respectively. While this will work, because the method signatures will be different, it's highly unintuitive if the users of our library do not know the netapi (and I think they shouldn't be expected to). 

So, a simple question in a short pull request: would you consider a different, more descriptive naming scheme for these methods, and shield the user of this library from the details of the salt netapi, or was this a deliberate decision?

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
@@ -147,7 +147,7 @@ public class SaltStackClient {
      * @return object representing the scheduled job
      * @throws SaltStackException if anything goes wrong
      */
-    public SaltStackJob minions(String target, String function, List<String> args,
+    public SaltStackJob startJob(String target, String function, List<String> args,
             Map<String, String> kwargs) throws SaltStackException {
         // Setup lowstate data to send as JSON
         JsonObject json = new JsonObject();
--- a/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
@@ -147,7 +147,7 @@ public class SaltStackClient {
      * @return object representing the scheduled job
      * @throws SaltStackException if anything goes wrong
      */
-    public SaltStackJob startJob(String target, String function, List<String> args,
+    public SaltStackJob startCommand(String target, String function, List<String> args,
             Map<String, String> kwargs) throws SaltStackException {
         // Setup lowstate data to send as JSON
         JsonObject json = new JsonObject();


isbm:
Incapsulation?

renner:
I would say it is OK to create a layer on top of the API endpoints, as long as we make sure that users of the library know which endpoint is used. It seems likely to me that the library will hold more than one method POSTing to `/minions`. The API actually expects [lowstate data] (http://docs.saltstack.com/en/latest/ref/netapi/all/salt.netapi.rest_tornado.html#term-lowstate) in the request body, which offers a lot of flexibility and our current `minions()` method is already a strong simplification of that!

We should therefore probably offer a `minions()` method that takes a `Lowstate` Java Object as parameter (doesn't exist yet) offering the full flexibility. We can even choose a different name, but `startJob()` doesn't fit here since you can actually run more than one command. There can be other methods talking to the same endpoint e.g. to just schedule the run of **one** command, like the current `minions()`. These are convenience methods and can have any appropriate name that should of course be chosen carefully (+ endpoint should be mentioned in JavaDoc).

About the naming itself my impression is that the docs talk more about *commands* than about *jobs*. So an appropriate name for the existing method might be *startCommand()* rather than *startJob()*?

If I understand correctly there could further be two variants of the GET version of /minions: `getMinions()` and `getMinionDetails(String mid)`.

isbm:
@renner Well, `/minions` is already convenience URL. :-) While I cannot agree more that this method is already simplification and sort of convenience method, I still think you're talking about a bit different case, where _convenience_ methods taking over the cumbersome burden of passing "raw" low-state parameters. In this case I would agree with you that methods, like `getMinions()` and `getMinionDetail(String mid)` or even  something like `getMinionDetail(Minion minion)`  would be a great _convenience additions_ to the main method, offered by Salt. This would indeed allow to us have a Java feel here and intuitively mapping this to the Salt. However, before that I would like to have an ability to call Salt with its way, something like passing previously constructed low state object.

But still I see nothing about "job" in the Salt documentation and in my opinion that would introduce something foreign.

renner:
Summarizing this into a general policy:

- We use the API endpoint name as the Java method name wherever it makes sense in order to make clear which endpoint a library method correlates to.
- In case the API endpoint name doesn't make sense we come up with appropriate method names that go along with the [netapi documentation] (http://docs.saltstack.com/en/latest/ref/netapi/all/salt.netapi.rest_cherrypy.html).

This documentation says about the `/minions` endpoint: 

    Start an execution command and immediately return the job id

Since `minions()` is clearly not a good name for the existing convenience method, would you mind to rename it to `startCommand()` in order to stick with that documentation as good as possible?

skhtmp:
Done, thanks for your comments!

moio:
Merging, thanks for your contributions!

PR #16, opened 2015-02-13
run-tests now runs all test in the test directory

lucidd:


renner:
Please provide a description for your PR. Maybe try to point out the advantages of the proposed patch, since I am personally not convinced yet that matching a wildcard is actually better than whitelisting all tests. What would happen for instance if there is a class called `FoobarTestUtils`, would you propose to exclude all `*Utils`?

You are not the first to propose this, see [my reply to another PR] (https://github.com/SUSE/saltstack-netapi-client-java/pull/11#commitcomment-9655170). So far I still tend to prefer a whitelist of tests, especially since there is only a small number of classes to list here. I know that `build.xml` has to be touched whenever there is a new test class, but how often do you expect that to happen?

lucidd:
Thanks for the feedback. You are right `**/*Test*` matches a bit too much that was actually not intended. I changed it now to only match classes that end on Test `**/*Test.java`. This approach enforces that all tests need to end with Test which i think is a good convention to follow.

The `build.xml` not only has to be changed when creating a new test but also needs to be kept in sync when refactoring and it will grow with each test case which leaves more room for mistakes. Since including a test is the more common case, it should not require more configuration than the more special case of excluding a test. It hurts more to accidentally forget to include a test then to forget to exclude it.

moio:
@lucidd, this second proposal makes sense, we discussed it with @renner and we agree this is the best short term solution.

Long term a port from `ant` to `maven` or some other build tool might make even more sense, but that's for another PR.

Merging, thanks!

PR #17, opened 2015-02-13
Implement POST /run

skhtmp:
As requested, my implementation of the run() method. I would prefer to call the method runCommand(), but am waiting for your comments on #15.

I'm not sure if my handling of return values is the best possible solution, it seems overkill to wrap them into even more tiny classes, but now the user needs to know which kind of composite Collection object is returned. I like it best as it is now, but your suggestions would be very welcome.


renner:
This method will execute one command in a synchronous way, so `runCommand()` could be a good name, as opposed to for instance `scheduleCommand()` for what we currently have as `minions()`. It is then just a bit unexpected that two methods with very similar names behave very differently regarding the session handling: scheduling a command will require to login first, while running a command needs the credentials every time. But maybe this is OK (as long as it is documented)?

The result type is IMO a bit too generic (`List<Map<String, String>>`), and why it even needs to be a list? It seems that this list always contains only one item, at least when looking at the example response, where there is results for several minions (`ms-0` through `ms-4`). I guess in this case `getResults()` could return `results.get(0)` to simplify this a bit, but we might want to verify that. It further needs to be documented that the map keys are the minion ids and the values are the result.

The patch looks quite good, please find some comments inline about minor things to fix. Thanks!

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -177,4 +178,56 @@ public SaltStackJob minions(String target, String function, List<String> args,
         // They return a list of tokens here, we take the first
         return result.getJobs().get(0);
     }
+    /**
```
renner: Please put one empty line between those methods.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -177,4 +178,56 @@ public SaltStackJob minions(String target, String function, List<String> args,
...
+        String client, String target, String function, List<String> args, Map<String,
+        String> kwargs) throws SaltStackException {
+
+        JsonObject json = new JsonObject();
+        json.addProperty("username", username);
+        json.addProperty("password", password);
+        json.addProperty("eauth",    eauth);
+        json.addProperty("client",   client);
+        json.addProperty("tgt",      target);
+        json.addProperty("fun",      function);
```
renner: Please don't make them aligned, instead use just one blank after every comma. What happens if there is later another property to be added that is even longer than *username*? All of these would need to be adjusted to be aligned again. This would blow up the patch (that should only add a property) or require a separate patch just to fix the alignment!

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -177,4 +178,56 @@ public SaltStackJob minions(String target, String function, List<String> args,
...
+     *
+     * @param username the username
+     * @param password the password
+     * @param eauth the eauth type
+     * @param client the client
+     * @param target the target
+     * @param function the function to execute
+     * @param args list of non-keyword arguments
+     * @param kwargs map containing keyword arguments
+     * @return object representing the scheduled job
```
renner: This is not true, it will actually return the results!

src/main/java/com/suse/saltstack/netapi/results/SaltStackRunResults.java
```diff
@@ -0,0 +1,16 @@
...
+import java.util.Map;
+
+import com.google.gson.annotations.SerializedName;
+
+public class SaltStackRunResults {
+
+    @SerializedName("return")
+    private List<Map<String, String>> results;
+
+    public List<Map<String, String>> getResults() {
```
renner: Please add javadoc comments at least for all public methods and the class itself. I know, existing result classes might not have that, but in here there is even a need to explain the results.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -177,4 +178,56 @@ public SaltStackJob minions(String target, String function, List<String> args,
...
+     * @throws SaltStackException if anything goes wrong
+     */
+    public SaltStackRunResults run(String username, String password, String eauth,
+        String client, String target, String function, List<String> args, Map<String,
+        String> kwargs) throws SaltStackException {
```
renner: Wrapped lines should be indented twice as much as the normal indentation = 8 spaces. Sorry, these rules are not yet written down anywhere in this repo.

skhtmp:
Thanks again for your comments. I will work on this one first and then return to #15

renner:
Good, but let's keep on using just `run()` as the name of this method, following the general naming policy from [#15] (https://github.com/SUSE/saltstack-netapi-client-java/pull/15#issuecomment-74485966).

skhtmp:
I checked against a running salt-api, and you are right: the list always has only one element, and the map has one entry per minion, so I handle it now exactly as minion() does.

If returning a Map is too generic, I can also wrap the result into a separate class, but would then propose SaltStackRunResultList and SaltStackRunResult (or even SaltStackResult{,List}?) and rename the wrapper class for the /minions response SaltStackJobList to be consistent.

As the netapi offers both /minions and /run I think this library should offer access to both endpoints, too. I could imagine that for short-running commands and when a bit of latency doesn't matter the synchronous, all-in-one-call run() method is convenient to have.

skhtmp:
(I see that this needs more work after #14 and #18, I'm working on it...)

skhtmp:
As the name SaltStackResult is already taken, I've renamed my return values to SaltStackReturnValues to reduce confusion. I think these should be reusable for parsing the results from /jobs too, once this is implemented.

SaltStackReturnValues is just a HashMap now. The keys are always Strings, but the type of the value Object is whatever Gson picks from the json string, and only the user of the method will know what these should be. If it's not a single value (as it is with test.ping) I'm not yet sure whether it's better to deserialize it here into whatever composite collection object, or better pass the json representation up to the caller.

So still work in progress, but I wanted to share what I have now.

renner:
I would say there is no need to introduce a new object that is exactly the same as HashMap, let's not do that. It was actually fine as it was before, so just keep the former `SaltStackRunResults` and return only the `Map<..>` instead of `List<Map<..>>` from the `getResults()` in there. It would even be OK to return this map directly from `run()` I guess, as you prefer.

skhtmp:
If I use the abstraction provided now by SaltStackParser, I really only need a HashMap to be filled by Gson. The previous construct with SaltStackRunResults having (instead of being) a Map containing the return values doesn't work any more, and I wanted to handle my json parsing the same as the other methods. So, one class less, SaltStackReturnValues is gone altogether.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -176,4 +177,59 @@ public SaltStackJob startCommand(String target, String function, List<String> ar
...
+            String client, String target, String function, List<String> args, Map<String,
+            String> kwargs) throws SaltStackException {
+
+        JsonObject json = new JsonObject();
+        json.addProperty("username", username);
+        json.addProperty("password", password);
+        json.addProperty("eauth", eauth);
+        json.addProperty("client", client);
+        json.addProperty("tgt", target);
+        json.addProperty("fun",      function);
```
lucidd: looks like an accidental indent :)

src/main/java/com/suse/saltstack/netapi/parser/SaltStackParser.java
```diff
@@ -25,6 +26,8 @@
         new SaltStackParser<>(new TypeToken<SaltStackResult<List<SaltStackToken>>>(){});
     public static final SaltStackParser<SaltStackResult<List<SaltStackJob>>> JOB =
         new SaltStackParser<>(new TypeToken<SaltStackResult<List<SaltStackJob>>>(){});
+    public static final SaltStackParser<SaltStackResult<List<HashMap>>> RETVALS =
+        new SaltStackParser<>(new TypeToken<SaltStackResult<List<HashMap>>>(){});
```
lucidd: I don't think we need it to be a `HashMap` so we can just use `Map` here.

renner: Yes, I agree, let's make it `Map<String, Object>` here and above. After that it can be merged (when the indent has been fixed as well ;)).

skhtmp: Done, thanks for your comments!

moio:
@skhtmp, thanks for this work and for the patience of going through multiple rounds of comments!

This looks good to me now, and since @renner's comments have all been dealt with I'm merging now :-)

skhtmp:
Thanks @moio!

PR #18, opened 2015-02-16
Generalisation of parser and result types + additional type safety using them.

lucidd:
The first commit generalizes the parser classes introduced in #14 to only require one class for all parsers instead of one per result type. It is still possible to create a subclass for each parser but it needs way less code now so its more boilerplate than "logic". Given the change in the parser it is now easy to have the connection classes return statically typed results based on the provided parser (done in the second commit).

The third commit is basically the same as the first one but for the `SaltStackResult` classes. (this might be adjusted or extended in the future depending on if there are results that follow a different structure).

Over all this PR results in less code for same functionality and adds additional type safety.

The only disadvantage i see so far with this is, that type annotations can become quite long. The `<>` notation helps a bit to reduce it but it is still a bit much. 
```java
SaltStackParser<SaltStackResult<List<SaltStackToken>>> TOKEN = new SaltStackParser<>(new TypeToken<SaltStackResult<List<SaltStackToken>>>(){});
```
One way to help with that would be dropping the `SaltStack` prefix on the classes since everything is already namespaced under the saltstack package anyway.

If you have any suggestions/feedback please let me know. :smiley:


src/main/java/com/suse/saltstack/netapi/parser/SaltStackParser.java
```diff
@@ -1,32 +1,39 @@
...
- * that apply to multiple results can be added to this class.
- */
-public abstract class SaltStackParser<T> {
-    public abstract T parse(InputStream inputStream) throws SaltStackParsingException;
+
+public class SaltStackParser<T> {
+
+    public static final SaltStackParser<SaltStackResult<String>> STRING = new SaltStackParser<>(new TypeToken<SaltStackResult<String>>(){});
+    public static final SaltStackParser<SaltStackResult<List<SaltStackToken>>> TOKEN = new SaltStackParser<>(new TypeToken<SaltStackResult<List<SaltStackToken>>>(){});
+    public static final SaltStackParser<SaltStackResult<List<SaltStackJob>>> JOB = new SaltStackParser<>(new TypeToken<SaltStackResult<List<SaltStackJob>>>(){});
```
isbm: Some wrapping here would be a good idea.

src/main/java/com/suse/saltstack/netapi/parser/SaltStackParser.java
```diff
@@ -1,32 +1,39 @@
...
+import java.util.List;
 
-/**
- * ABS for result parsers. All result parsers need to inherit from this class. Generic methods
- * that apply to multiple results can be added to this class.
- */
-public abstract class SaltStackParser<T> {
-    public abstract T parse(InputStream inputStream) throws SaltStackParsingException;
+
+public class SaltStackParser<T> {
```
renner: Please add javadoc comments at least to all classes as well as to all public methods.

src/main/java/com/suse/saltstack/netapi/parser/SaltStackParser.java
renner: Yes, please wrap long lines. It's not (yet) written down somewhere in this repo, but I tend to like a maximum line length of 92 characters. Please apply this to other long lines as well.

moio:
>One way to help with that would be dropping the SaltStack prefix on the classes since everything is already namespaced under the saltstack package anyway.

Right, but please in a different PR :-)

lucidd:
@renner  I think it should be fine now if there is still something missing please let me know.
@moio I didn't plan on including it in this PR anyway since it was just a suggestion but if dropping the `SaltStack` prefix is a desired change i will gladly prepare a PR for it.

moio:
Merged, thanks for your contribution!

Of course new PRs are always welcome!

PR #20, opened 2015-02-16
Make working with the SaltStackClientConfig more typesafe

lucidd:
This replaces the current stringly typed `SaltStackClientConfig` with a statically typed key/value store that knows the type of each key. This shifts the burden of parsing the configuration values from the consumers side to the producers side. So everything querying the config for values does not have to deal with parsing and error handling of invalid values like `"hello"` as a port value. This also results in those invalid values being detected earlier (for example when the config is loaded from a file instead of when a value is used).

I would also suggest to generally make parameters stricter and ask for what is really needed instead of something that could potentially be parsed to what is needed. An example of this is the change [here](https://github.com/lucidd/saltstack-netapi-client-java/compare/config#diff-250997581e435e368654d6b84ccbe400R52) asking directly for a URI object instead of a String.
This reduces the overall error handling needed inside the API and and shifts it to the user of the API which should be fine since most argument parsing errors would probably result in an error that bubbles up to the user anyway.

I also experimented with a version that can prevent different sub types of `KeyValueStore` from mixing keys with each other but left it out for now since there is only one Configuration class right now. But it could be useful in the future.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
@@ -1,6 +1,7 @@
 package com.suse.saltstack.netapi.client;
 
 import com.suse.saltstack.netapi.config.SaltStackClientConfig;
+import static com.suse.saltstack.netapi.config.SaltStackClientConfig.*;
 import com.suse.saltstack.netapi.config.SaltStackProxySettings;
 import com.suse.saltstack.netapi.exception.SaltStackException;
 import com.suse.saltstack.netapi.parser.SaltStackParser;
@@ -12,7 +13,7 @@ import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonPrimitive;
 
-import java.net.URISyntaxException;
+import java.net.URI;
 import java.util.List;
 import java.util.Map;
 
@@ -33,7 +34,7 @@ public class SaltStackClient {
      * @param url the SaltStack URL
      * @throws SaltStackException
      */
-    public SaltStackClient(String url) throws SaltStackException {
+    public SaltStackClient(URI url) {
         this(url, new SaltStackHttpClientConnectionFactory());
     }
 
@@ -44,16 +45,9 @@ public class SaltStackClient {
      * @param connectionFactory Connection Factory implementation
      * @throws SaltStackException
      */
-    public SaltStackClient(String url, SaltStackConnectionFactory connectionFactory)
-            throws SaltStackException {
+    public SaltStackClient(URI url, SaltStackConnectionFactory connectionFactory) {
         // Put the URL in the config
-        if (url != null) {
-            try {
-                config.setUrl(url);
-            } catch (URISyntaxException e) {
-                throw new SaltStackException(e);
-            }
-        }
+        config.put(URL, url);
         this.connectionFactory = connectionFactory;
     }
 
@@ -73,13 +67,13 @@ public class SaltStackClient {
      */
     public void setProxy(SaltStackProxySettings settings) {
         if (settings.getHostname() != null) {
-            config.put(SaltStackClientConfig.PROXY_HOSTNAME, settings.getHostname());
-            config.put(SaltStackClientConfig.PROXY_PORT, String.valueOf(settings.getPort()));
+            config.put(PROXY_HOSTNAME, settings.getHostname());
+            config.put(PROXY_PORT, settings.getPort());
         }
         if (settings.getUsername() != null) {
-            config.put(SaltStackClientConfig.PROXY_USERNAME, settings.getUsername());
+            config.put(PROXY_USERNAME, settings.getUsername());
             if (settings.getPassword() != null) {
-                config.put(SaltStackClientConfig.PROXY_PASSWORD, settings.getPassword());
+                config.put(PROXY_PASSWORD, settings.getPassword());
             }
         }
     }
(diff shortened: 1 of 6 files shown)


moio:
We discussed this with @isbm and @renner and generally like the underlying ideas, thanks for submitting this PR.

What we currently don't like is the design of classes, specifically, the fact that `KeyValueStore` is separate and abstract. After your changes it is only used to implement `SaltStackClientConfig`, which then becomes a mere collection of constants, moreover not implementing any abstract method.

We think it would be better to merge `KeyValueStore` code into `SaltStackClientConfig` to keep design simple, at least until we have an actual need of reusing `KeyValueStore` somewhere else. That way `SaltStackClientConfig` will have its inner `Key` class with default value, which sounds correct, and will internally use `HashMap<Key, Object>` as an implementation detail.

What do you think?

lucidd:
The design mostly comes from my experiments with multiple sub classes and a second type parameter to prevent mixing keys. You are right that this does not make much sense if we only have one sub class.
I will merge those too together and rebase this PR onto the latest master so it is mergeable without conflicts. 

Out of curiosity, what would you do differently in regards to sub classes only being used as a namespace for keys, if there were multiple sub classes of `KeyValueStore`? I personally don't mind using them that way but i'm open for suggestions.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -44,15 +45,11 @@ public SaltStackClient(String url) throws SaltStackException {
      * @param connectionFactory Connection Factory implementation
      * @throws SaltStackException
      */
-    public SaltStackClient(String url, SaltStackConnectionFactory connectionFactory)
+    public SaltStackClient(URI url, SaltStackConnectionFactory connectionFactory)
             throws SaltStackException {
```
moio: Looks like `SaltStackException` is not thrown anymore, please cleanup method signature and any relevant callers.

src/main/java/com/suse/saltstack/netapi/config/SaltStackClientConfig.java
```diff
@@ -1,132 +1,92 @@
...
-    }
-
-     /**
-     * Returns the configured proxy port (or 3128 as the default).
-     *
-     * @return proxy port
-     */
-    public String getProxyPort() {
-        return properties.getProperty(PROXY_PORT, DEFAULT_PROXY_PORT);
+    @SuppressWarnings("unchecked")
```
moio: At least in Java 7, this `SuppressWarnings` seems unnecessary. Is this needed in Java 8?

src/main/java/com/suse/saltstack/netapi/config/SaltStackClientConfig.java
```diff
@@ -1,132 +1,92 @@
...
-    /**
-     * Returns the session token or null.
-     *
-     * @return session token
-     */
-    public String getToken() {
-       return properties.getProperty(TOKEN, null);
-    }
 }
+
```
moio: Blank like probably forgotten here.

moio:
> what would you do differently in regards to sub classes only being used as a namespace for keys, if there were multiple sub classes of `KeyValueStore`? I personally don't mind using them that way but i'm open for suggestions.

I am not sure I understand your question - are you wondering what to do with inner classes like `KeyValueStore.Key` when the outer class is subclassed?

In that case, I would evaluate how strong the inner and outer class "link" is, semantically. If it is not too strong, I would simply move the inner class in a new separate one - after all inner classes are supposed to be so highly coupled with their outer ones to be basically part of them.

src/main/java/com/suse/saltstack/netapi/config/SaltStackClientConfig.java
lucidd: You are right its no longer needed in this implementation. A previous one did use some casting.
Using Java 8 didn't produce any warnings about it either.
Speaking of Java 8 is there any chance of upgrading to is or is 7 a hard requirement.
I'm just asking because it would be nice to use `Optional` instead of passing around `null` everywhere.

moio: At the moment it is a hard requirement. Rest assured that myself and others are hoping and pushing for it to be relaxed as soon as possible!

lucidd:
The question was more about `SaltStackClientConfig` being just a collection of constants and not having anything else. If we had the super type `KeyValueStore` and two sub classes `A` and `B` that each contain only their key constants would that be okay or would there be a better way of doing this?

lucidd:
Ok i think it should be fine now i addressed all the points you mentioned.

src/main/java/com/suse/saltstack/netapi/config/SaltStackClientConfig.java
lucidd: Thats good to here. Coming from scala i really miss that stuff and Java 8 is definitely a step in the right direction in that regard :smile:

moio:
> If we had the super type `KeyValueStore` and two sub classes `A` and `B` that each contain only their key constants would that be okay or would there be a better way of doing this?

OK, I understand now. I don't think using classes as mere constant containers is a great design idea in general, it certainly doesn't sound very object-oriented. Depending on the situation I would probably try to design classes differently so I can stick constants to related classes that also perform some kind of function. That said the idea might prove valuable if a very high number of constants is needed for some reason, even if I haven't stumbled in such a case yet.

PR is good, merging, thanks for your contributions so far.

PR #21, opened 2015-02-17
Make parsers static to avoid unnecessary allocations since they're stateless

dgiagio:
At every request a parser is instantiated to handle the response. Since the parsers are stateless this patch instantiates them only once, reusing the same object for different responses.


moio:
Good point and good patch, unfortunately basically equivalent code is included in #18 which has been merged already.

Sorry and thanks for your contribution anyway!

PR #23, opened 2015-02-18
Constant, Exception, logging and validation updates on SaltStackClient.java

buraksarac:
I have just made few changes on SaltStackClient.java
1. Client constructor overloaded to handle port, host and ssl
2. Validation added into client constructor
3. Logger instance added
4. String null checks replaced with apache StringUtils.isEmpty()
5. Predefined exceptions added
6. String password changed to a Password object #see http://securesoftware.blogspot.cz/2009/01/java-security-why-not-to-use-string.html
7. All hardcoded strings caried into api constants
8. Exception handling improved


moio:
@buraksarac, good morning and thanks for contributing!

I would like to merge at least part of your work, but before doing that it would be good if you resubmitted it in an easier-to-review form. Here are some guidelines, which by the way are also valuable in most open source projects and particularly ones in GitHub:

 - open a PR for each clearly separated topic. Smaller PRs are easier to check, can be looked at in parallel by multiple reviewers and tend to result in less merge conflicts, so it's more likely they get merged sooner. Plus any comments from our part that "block" one of them would not slow down the merging of others!
 - every PR should be both described and motivated in comments. It's good to have a one-line overview of what you have done but we can also simply read your code - much more important is why you did that, and that is not easy to read from Java source!
 - if your PR `B` depends on another PR `A`, include commits from the `A` branch in `B`. You will still have to rework `B` in case `A` does not get accepted as-is, but at least it will be clear to the reviewers that this dependency exists. Obviously try to work on independent PRs as much as possible;
 - if possible, retain meaningful commits in your PR and don't squash all changes in one. There might be value in the history of how you got to a particular solution, and merge commits will retain a track of the whole PR change anyway;
 - adding new dependencies (libraries) should be at least motivated a bit. Is it worth it to add a new dependency? Why did you choose a particular library? Were there any alternatives, why do you think your choice makes sense?
 - don't change indentation of lines you did not otherwise edit, unless your contribution is a whitespace cleanup - it makes it hard to understand what your contribution is ([this is what we see](https://github.com/SUSE/saltstack-netapi-client-java/pull/23/files)); 
 - in this project's particular case, we indent with 4 spaces and do not use tabs ([as specified in the README](https://github.com/SUSE/saltstack-netapi-client-java));
 - same goes for reorganizing imports. Please do not change the ordering unless your contribution is exactly about that.

I look forward to see comments, and further contributions, from you :-)

buraksarac:
Hi @moio goodmorning and thank you for comments!

  I am going to separate PRs today with more explanation. I think job notice confuse me a bit so I have created PR at once as a reference because I didnt wanted to flood:)
  New dependencies: Apache utils are quite powerfull especially on collection and String operations. I have added those libraries because beside of null check it is also doing an empty check. Instead of writing conditional check per case I thought it would be a  good shot to give credit to another open source project. 
 What IDE are you (guys) using, if it is eclipse can you provide me your formatter template or include into project folder? If not I will figure out:)
 Have you ever considered using maven? Since it is opensource and early stage project can be separated as multi-maven project (api and implementation) so if someone else wants to do a implementation can use api.jar? Maven also can help for dependency management from a centralised repo so you will not need to do hash check or you can run your test by Jenkins nightly builds using goals instead every buid. If you are positive I can volunteer for implementation:)
 Thanks
 Burak

renner:
Hey, I am personally using eclipse, but everybody is of course free to work with whatever they prefer. We will therefore not provide a formatter configuration for a specific IDE in the repo. For you it is still easy to figure out how to setup yours: please find the most important settings [here] (https://github.com/SUSE/saltstack-netapi-client-java).

Generally we would be happy to see a pull request adding support for maven builds, please feel encouraged to work on that. I doubt that the concept of multiple modules as you are suggesting it is applicable for this project though. I am not 100% sure if I understood you correctly, but I don't think there is currently much value being added by splitting up this library into two separate jars, for api and implementation respectively.

buraksarac:
Hello @renner thanks for the input. I have setted up Eclipse it was ok, now I am separeting PRs. I think I will go step by step:) I already have one maven version as a single project, I wanted to see yesterday how it will work. So later I can put into my repo for your review.

PR #24, opened 2015-02-19
Dropping the SaltStack prefix [WIP]

lucidd:
This PR is still work in progress.

The individual renames are in separate commits to make reviewing easier and help git to keep track of the renames better since some files consist mostly of type names.

#### Motivation:

As initially proposed in #18 dropping the `SaltStack` prefix reduces the line length in a lot of cases which makes it easier to obey the 92 limit without wrapping too often. The intend of the classes should still be clear since they are all name spaced with the `com.suse.saltstack` package.

#### Done so far:

* Dropping the `SaltStack` for all classes except `SaltStackParser` and `SaltStackException`.
* `SaltStackParser` got renamed to `JsonParser` to be more descriptive.

#### Needs discussion:

* `SaltStackException` has not been renamed yet since just `Exception` somehow felt wrong but its probably the right thing to do.
* `Token` might be a bit to generic one alternative would be `AuthToken`.

Let me know what you think. Any alternative naming ideas are welcome.

#### Additional ideas for followup PRs:

* Rename and move `APIConstants` to `com.suse.saltstack.netapi.Constants`
since it contains the constants of the saltstack netapi.
* Moving `com.suse.saltstack.netapi.results.{Job, Token}` to `com.suse.saltstack.netapi.datatypes` since those represent datatypes of the saltstack netapi.
* Move the `Connection` and `ConnectionFactory` implementation into a `impl` sub package.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/client/APIConstants.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/APIConstants.java
@@ -3,7 +3,7 @@ package com.suse.saltstack.netapi.client;
 /**
  * Constants used in SaltStack API.
  */
-public class SaltStackAPIConstants {
+public class APIConstants {
     /** Known values for 'eauth' parameter when logging in */
     public static final String LOGIN_EAUTH_AUTO = "auto";
     public static final String LOGIN_EAUTH_PAM = "pam";
--- a/src/main/java/com/suse/saltstack/netapi/client/Connection.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/Connection.java
@@ -7,7 +7,7 @@ import java.lang.reflect.Type;
 /**
  * Describes an interface for different HTTP connection implementations.
  */
-public interface SaltStackConnection<T> {
+public interface Connection<T> {
     /**
      * Send a POST request and parse the result into object of given {@link Type}.
      *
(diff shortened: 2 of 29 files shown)


moio:
SaltStackException: I would leave it as it is.

Token: I would call it Token, since that is [how the official API describes it](http://docs.saltstack.com/en/latest/topics/eauth/index.html#tokens), see also [the discussion over naming](https://github.com/SUSE/saltstack-netapi-client-java/pull/15#issuecomment-74485966).

APIConstants: I agree with your proposal.

It's mostly a matter of taste, thus I happily leave the last word to @renner :-)

renner:
I agree with @moio regarding `SaltStackException` and `Token`, both are good choices. There is one rename in this PR that I would suggest to revert, and that is `SaltStackClient` -> `Client`. That class is the main entry point of this library, so my personal taste clearly prefers this:

    SaltStackClient client = new SaltStackClient(...);

A reader of such code knows immediately what is going to happen, while just `Client` is a bit too generic IMO. Apart from that I like the changes and we can merge.

I also like the suggested further package refactorings, thanks for your efforts!

lucidd:
@moio @renner Thanks for the feedback i removed the `SaltStackClient` rename and brought the patch up to date with the master branch. 

Since this PR is already big enough i will submit the other package refactorings in a separate PR. :smile:

moio:
Merging, thanks!

PR #25, opened 2015-02-19
Ignore all .iml files

lucidd:
Intellij also puts `.iml` files in sub directories if they are configured as modules (for example having `main` and `test` configures as separate modules).


renner:
Thanks, merging.

PR #26, opened 2015-02-19
Add 'start' and 'expire' dates to token

hustodemon:
This patch adds dates to login token. For parsing dates, new
deserializer was added to SaltStackParser class.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
+++ b/src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
@@ -1,6 +1,11 @@
 package com.suse.saltstack.netapi.parser;
 
 import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonParseException;
 import com.google.gson.reflect.TypeToken;
 import com.suse.saltstack.netapi.results.*;
 
@@ -8,6 +13,8 @@ import java.io.BufferedReader;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
+import java.lang.reflect.Type;
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/suse/saltstack/netapi/parser/SaltStackParser.java
```diff
@@ -50,6 +60,25 @@ public T parse(InputStream inputStream) throws SaltStackParsingException {
...
+    private class SaltStackDateDeserializer implements JsonDeserializer<Date> {
+        @Override
+        public Date deserialize(JsonElement jsonElement, Type type,
+                JsonDeserializationContext jsonDeserializationContext)
+                throws JsonParseException {
+            try {
+                double dateMiliSecs = jsonElement.getAsDouble() * 1000;
+                return new Date((long) dateMiliSecs);
+            } catch (Exception e) {
+                e.printStackTrace(); // TODO ideally use some logging infrastructure
```
moio: Please throw an appropriate `JsonParseException` here in case the format is not as expected and do not swallow other unexpected exceptions.

moio:
Hey @hustodemon, thanks for this pull request!

We like the idea and the patches, apart from two minor points:

 - in `SaltStackDateDeserializer`, throw proper exception (see line comment);
 - since PR #24 got merged before this one, please rename files/classes so that merge is clean again. Sorry for the double work!

Looking forward to hear comments from you!

renner:
I tested the changes and it works fine, very nice to have those login token objects completed. 03de503eca998f03a12ab5f0331b2b2bed5a78cf is actually an independent change that could even have gone into a separate PR, right? Nevermind though, since this PR is not very big overall we can handle it in here.

Maybe let's move that `makeJsonData()` method away into a utils class though, or at least to the end of that class file in order to keep the public methods together?

src/main/java/com/suse/saltstack/netapi/parser/SaltStackParser.java
hustodemon: Sure, I'll do it!

hustodemon:
Hi guys,

thanks for your comments!
@renner I'll split the patches for sure. Pushing the second change here was just an accident :)

Cheers.

renner:
I merged it, thanks for your contribution!

hustodemon:
Thanks!

PR #27, opened 2015-02-19
Code and Javadoc cleanup

dgiagio:
- Code cleanup such as removing extra semicolon, redundant or unused variables, incorrect exception throw specifications and unnecessary boxing for primitive integer type.
- Add missing Javadoc and remove incorrect ones.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
@@ -32,7 +32,6 @@ public class SaltStackClient {
      * Constructor for connecting to a given URL.
      *
      * @param url the SaltStack URL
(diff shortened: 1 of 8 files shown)


lucidd:
Haha seems like we had the same idea but since you were faster can you include the deletion of [this](https://github.com/SUSE/saltstack-netapi-client-java/blob/master/src/main/java/com/suse/saltstack/netapi/client/SaltStackJDKConnection.java#L51) documentation of an outdated parameter. :smile:

dgiagio:
@lucidd This is already in the PR. :-)

lucidd:
Oh you're right my bad didn't see it at first. :)

moio:
Merge-Just-Do-It (TM).

Thanks a lot.

PR #28, opened 2015-02-19
Add asynchronous support for all implemented requests

dgiagio:
- Add asynchronous support for all implemented requests. Allows one to specify an Executor but defaults to Executors.newCachedThreadPool() if none is given.
- Java 7 compatible (not using Java 8 new concurrent features).
- Add testing for asynchronous requests. ```ant run-tests``` clean.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
@@ -16,6 +16,10 @@ import com.google.gson.JsonPrimitive;
 import java.net.URI;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 
 /**
  * SaltStack API client.
@@ -27,6 +31,9 @@ public class SaltStackClient {
 
     /** The connection factory object */
     private ConnectionFactory connectionFactory;
+
+    /** The executor for async operations */
+    private ExecutorService executor;
 
     /**
      * Constructor for connecting to a given URL.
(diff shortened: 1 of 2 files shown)


lucidd:
Nice work! Just one thing do you think this Future wrapper is necessary? I think i would be better to just return Future and have utility functions to work with them. That way we don't loose the ability to pass them to other functions and libraries expecting Future.

moio:
@dgiagio: very cool, and thanks for contributing again!

I tend to agree with @lucidd here -  current implementation hides potentially useful methods in `Future` such as `isDone`. Those might or might not be useful depending on the threading model of the caller - but I tend to favor the fact we don't assume any.

I understand that "automatic unboxing" of SaltStackExceptions is handy, but it also makes it more difficult for library clients to handle concurrency-specific exceptions such as `InterruptedException` in some special way if they wanted to.

Bottom line FMPOV: let's not lose generality and use `Future` directly. @renner, opinions?

dgiagio-druid:
@lucidd and @moio I considered using Future<T> directly, but that imposes some interface "dificulties" to users, such as can't throw our own exceptions in get() methods (Future<T> throws InterruptedException and ExecutorException). The "real" exception thrown by get() would be: ExecutorException if there's some kind of exception internal to the Executor itself (e.g. out of threads for running the task) or caused by the Callable<T> itself which is wrapped as the cause (which might be SaltStackException). The boiler plate to write to make things work that way is unpleasant, which made me decide in creating this wrapper class. As far as I know, that is common practice.

About methods such as isDone(), we can implement them in our own SaltStackFuture delegating the implementation to the wrapped future. About the InterruptedException, we can choose to throw that from our own get() methods if you wish.

What do you think?

moio:
@dgiagio, IMHO users of the async interface are already a bit more "advanced" and will likely have bigger problems than catching a couple of exceptions properly. For quick hacks, the synchronous interface is simple, if you want to use threads then you are doing serious work and it's justified that the interface is a bit more complex.

Creating our own similar-but-slightly-incompatible version of Future is out of scope for this project, we better stick to the JDK-provided one which is standard and hopefully wisely designed by people at Oracle (we don't want to take their job!).

I talked with @renner about this and we agree.

dgiagio:
@moio @renner Fair enough. I will change the implementation to use Future<T> directly. Thanks for the input.

lucidd:
@dgiagio I think the big advantage of using the standard JDK future is that if we or someone using this project pulls in a library for working with futures it will be easier for them to interoperate with our API.
I agree that working with Javas Future is currently not the best experience compared to other JVM languages but that will hopefully improve over time.

The error handling you included in your wrapper could still be provided with helper methods in something like a `FutureUtils`. That said i'm personally not a big fan of wrapping Exceptions just for the sake of wrapping them since i feel like it hides away that fact that there are different things that can go wrong that should be handled separately. The only reason i can see this being useful is if you can provide additional information with the wrapper exception with is not the case here.

dgiagio:
@lucidd Yes, agreed. Like @moio said, users of the async interface probably know how to deal with Future<T> already, so it's not really necessary to wrap it. The PR has been modified to account for that. Thanks for all input!

moio:
@renner, I am OK with this pull request. If you also agree, please merge.

@dgiagio, thanks again!

renner:
Sorry that it took me so long: I just tested it and I like the patch, thanks @dgiagio! It works fine as well, so I will merge it now. I wonder if it would make sense to move those `*Async` methods away into a separate namespace, i.e. a special `SaltStackAsyncClient` class? Otherwise the `SaltStackClient` class could grow quite big once we implemented support (+ async support) for more API endpoints. Further only the async version of the client needs an `ExecutorService`.

PR #29, opened 2015-02-20
*LogUtils added for logging

buraksarac:
Hello All,
  I have created **LogUtils.java** to handle logger instances. So since now you will not need to create any logger instance in each class you wrote. Instead you can simply call i.e. `LogUtils.debug("hello salt");`. There is also predefined methods you can use in constructor, start or at the end of the method.
 **How does it work:** Simply collects caller class and method name by its thread, then creates a Slf4j logger instance using this class name and appends method name into starting of message. While creating logger instance its adding into local cache so in your next call you will received already created instance.
 **Usage:** Update *src/(main|test)/resources/simplelogger.properties* file. i.e. to receive trace messages from classes under `com.suse.*` package set `org.slf4j.simpleLogger.log.com.suse=warn`
**Files changed:** *src/test/resources* folder included into ant build so I have updated build.xml, I have updated *SaltStackClient.java* using this utility class, 'SaltStackProxySettings#toString()' overriden. I have also updated build.xml to see failure trace(if there is any).

**Simple output from build:***
    [junit] Running com.suse.saltstack.netapi.client.SaltStackClientTest
    [junit] Testsuite: com.suse.saltstack.netapi.client.SaltStackClientTest
    [junit] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.446 sec
    [junit] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.446 sec
    [junit] ------------- Standard Error -----------------
    [junit] 2015-02-20 01:43:47:627 +0100 [main] DEBUG com.suse.saltstack.netapi.client.SaltStackClient - <init>() constructed
    [junit] 2015-02-20 01:43:48:091 +0100 [main] DEBUG com.suse.saltstack.netapi.client.SaltStackClient - run() start. Username: user
    [junit] 2015-02-20 01:43:48:092 +0100 [main] TRACE com.suse.saltstack.netapi.client.SaltStackClient - run() Adding credentials into json. Target: *,client:local,function: test.ping*

*p.s. I will check your response tomorrow and do changes if you ask any but for other futures I am not sure if I can support this week I am going to join a competition which I am going to lose:)*

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -27,7 +27,7 @@
 
   <!-- Classpath for building and running tests -->
   <path id="classpath-tests">
-  	<pathelement location="src/test/resources"/>
+    <pathelement location="src/test/resources"/>
     <path refid="classpath-build" />
     <fileset dir="${dir-dist}">
       <include name="**/*.jar" />
--- a/build.xml
+++ b/build.xml
@@ -27,6 +27,7 @@
 
   <!-- Classpath for building and running tests -->
   <path id="classpath-tests">
+  	<pathelement location="src/test/resources"/>
     <path refid="classpath-build" />
     <fileset dir="${dir-dist}">
       <include name="**/*.jar" />
@@ -104,7 +105,7 @@
               <include name="**/*Test.java" />
           </fileset>
       </batchtest>
-      <formatter type="plain"/>
+      <formatter type="plain" usefile="false"/>
     </junit>
   </target>
 
--- a/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
@@ -20,7 +20,6 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
-
 /**
  * SaltStack API client.
  */
@@ -71,7 +70,7 @@ public class SaltStackClient {
      * @param settings proxy settings
      */
     public void setProxy(SaltStackProxySettings settings) {
-    	LogUtils.debug("start settings: " + settings.toString());
+        LogUtils.debug("start settings: " + settings.toString());
         if (settings.getHostname() != null) {
             config.put(PROXY_HOSTNAME, settings.getHostname());
             config.put(PROXY_PORT, settings.getPort());
@@ -108,8 +107,8 @@ public class SaltStackClient {
      */
     public SaltStackToken login(String username, String password, String eauth)
             throws SaltStackException {
-    	LogUtils.debug("start username:" + username);
-    	LogUtils.trace("Adding json properties");
+        LogUtils.debug("start username:" + username);
+        LogUtils.trace("Adding json properties");
         JsonObject json = new JsonObject();
         json.addProperty("username", username);
         json.addProperty("password", password);
@@ -132,7 +131,7 @@ public class SaltStackClient {
      * @throws SaltStackException if anything goes wrong
      */
     public SaltStackResult<String> logout() throws SaltStackException {
-    	LogUtils.debugStart();
+        LogUtils.debugStart();
         SaltStackResult<String> result = connectionFactory
                 .create("/logout", SaltStackParser.STRING, config).getResult(null);
         config.remove(TOKEN);
@@ -154,15 +153,15 @@ public class SaltStackClient {
      */
     public SaltStackJob startCommand(String target, String function, List<String> args,
             Map<String, String> kwargs) throws SaltStackException {
-    	LogUtils.debug("start. Target: " + target + " ,function: " + function);
+        LogUtils.debug("start. Target: " + target + " ,function: " + function);
         // Setup lowstate data to send as JSON
-    	LogUtils.trace("Adding json properties");
+        LogUtils.trace("Adding json properties");
         JsonObject json = new JsonObject();
         json.addProperty("tgt", target);
         json.addProperty("fun", function);
         // Non-keyword arguments
         if (args != null) {
-        	LogUtils.trace("Adding arguments into json: " + Arrays.toString(args.toArray()));
+            LogUtils.trace("Adding arguments into json: " + Arrays.toString(args.toArray()));
             JsonArray argsArray = new JsonArray();
             for (String arg : args) {
                 argsArray.add(new JsonPrimitive(arg));
@@ -171,7 +170,7 @@ public class SaltStackClient {
         }
         // Keyword arguments
         if (kwargs != null) {
-        	LogUtils.trace("Adding kwargs into json");
+            LogUtils.trace("Adding kwargs into json");
             for (String key : kwargs.keySet()) {
                 json.addProperty(key, kwargs.get(key));
             }
@@ -181,7 +180,8 @@ public class SaltStackClient {
 
         // Connect to the minions endpoint and send the above lowstate data
         SaltStackResult<List<SaltStackJob>> result = connectionFactory
-                .create("/minions", SaltStackParser.JOB,  config).getResult(jsonArray.toString());
+                .create("/minions", SaltStackParser.JOB, config).getResult(
+                        jsonArray.toString());
 
         LogUtils.debugEnd();
         // They return a list of tokens here, we take the first
@@ -204,12 +204,12 @@ public class SaltStackClient {
      * @return Map key: minion id, value: command result from that minion
      * @throws SaltStackException if anything goes wrong
      */
-    public Map<String,Object> run(String username, String password, String eauth,
+    public Map<String, Object> run(String username, String password, String eauth,
             String client, String target, String function, List<String> args, Map<String,
             String> kwargs) throws SaltStackException {
-    	LogUtils.debug("start. Username: " + username);
-    	LogUtils.trace("Adding credentials into json. Target: " + target + ",client:"
-		        + client + ",function: " + function);
+        LogUtils.debug("start. Username: " + username);
+        LogUtils.trace("Adding credentials into json. Target: " + target + ",client:"
+                + client + ",function: " + function);
         JsonObject json = new JsonObject();
         json.addProperty("username", username);
         json.addProperty("password", password);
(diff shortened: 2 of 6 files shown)


build.xml
```diff
@@ -27,6 +27,7 @@
 
   <!-- Classpath for building and running tests -->
   <path id="classpath-tests">
+  	<pathelement location="src/test/resources"/>
```
moio: Tab here, please always use spaces!

buraksarac:
@moio I think all files updated, I have disabled mixed mode so just spaces currently.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -1,19 +1,22 @@
 package com.suse.saltstack.netapi.client;
 
 import com.suse.saltstack.netapi.config.SaltStackClientConfig;
+
```
moio: Empty lines added here.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -165,8 +180,10 @@ public SaltStackJob startCommand(String target, String function, List<String> ar
 
         // Connect to the minions endpoint and send the above lowstate data
         SaltStackResult<List<SaltStackJob>> result = connectionFactory
-                .create("/minions", SaltStackParser.JOB,  config).getResult(jsonArray.toString());
+                .create("/minions", SaltStackParser.JOB, config).getResult(
+                        jsonArray.toString());
```
moio: This line has been rewrapped. Please only submit changes for lines that are meant to be changed!

src/main/java/com/suse/saltstack/netapi/config/SaltStackProxySettings.java
```diff
@@ -93,4 +93,17 @@ public String getPassword() {
     public void setPassword(String password) {
         this.password = password;
     }
+
+    @Override
+    public String toString() {
```
moio: Public methods should have a comment, please add one here.

src/main/resources/simplelogger.properties
```diff
@@ -0,0 +1,34 @@
...
+# Defaults to true.
+#org.slf4j.simpleLogger.showThreadName=true
+
+# Set to true if you want the Logger instance name to be included in output messages.
+# Defaults to true.
+#org.slf4j.simpleLogger.showLogName=true
+
+# Set to true if you want the last component of the name to be included in output messages.
+# Defaults to false.
+org.slf4j.simpleLogger.showShortLogName=false
```
moio: No newlines at end-of-file, please add.

build.xml
```diff
@@ -27,6 +27,7 @@
 
   <!-- Classpath for building and running tests -->
   <path id="classpath-tests">
+    <pathelement location="src/test/resources"/>
```
moio: I am currently wondering why those changes are needed, as this PR does not contain any change in test code. Can you explain please?

src/main/java/com/suse/saltstack/netapi/utils/LogUtils.java
```diff
@@ -0,0 +1,297 @@
...
+
+/**
+ * Utility class for handling logger instances and printing parsed messages according to caller method<br/>
+ * Logger instances will be created using reflection but the performance effect will be minimum since<br/>
+ * instances handled by local cache. <br/>
+ * <br/>
+ * <b>Usage:</b><br/>
+ *       In your class simply call <b><code>LoggingUtils.debug("hello logger");</code></b><br/>
+ * Simply edit log4j.properties file from project folder to handle logging level.<br/>
+ * @author Burak Sarac, burak at linux dot com, 20/02/2015 00:47
```
moio: @renner: do we want authorship and license notices in comments? Personally I think they are not very useful, since `git blame` already does that in a more precise way.

src/main/java/com/suse/saltstack/netapi/utils/LogUtils.java
```diff
@@ -0,0 +1,297 @@
...
+ * 
+ * License: Free of use!
```
moio: This is a bit too generic. I expect your contribution follows [the project LICENSE](https://github.com/SUSE/saltstack-netapi-client-java/blob/master/LICENSE), otherwise it can't be accepted of course.

moio:
OK @buraksarac, thanks for this further submission! We are definitely getting closer to a merge :-)

Some points though.

First, there is a number of smallish details that I outlined in inline comments, please fix those.

Second, there are some bigger points I am not sure about. I expect colleagues and other contributors to chime in to form an opinion, of course also feel free to comment yourself:
 - do we need that fine-grained level of logging in `SaltStackClient`?
 - will we accept this patch, even if the new class `LoggingUtils` comes with no tests?
 - while the idea of using reflection to find out classes is nice as it avoids initialization code, I am concerned that we are limiting the original API with the `LoggingUtils` facade a bit too much here. For example, one of the nice things in SLF4J is that it will automatically skip string generation even without using if-guards. This code:

    logger.debug("Temperature set to {}. Old temperature was {}.", t, oldT);

will not result in any string concatenation or `toString` call if the debugging level is above `DEBUG`. With this PR we totally lose this capability;
 - related to the above, I am not sure that completely wrapping SLF4J is the right design choice here. We are already replicating debug level flags here, we might end up duplicating a lot of stuff as we expose more functionality.

Third, it looks like some of the earlier PRs conflict, so please take a look and rebase if necessary. I am sorry for this extra work, but unfortunately we have no way to avoid it when we get multiple submissions at the same time!

lucidd:
I definitely like the idea. I'm a big fan of reducing boilerplate code. That said here are a few improvements and concerns i have.

* I think the traces in between entry and exit of a method are a little bit to much. Logging only the method entry with all the non confidential parameters should be enough information and keeps the code cleaner.

* The lazy string generation can easily be done by passing through the varargs to SLF4J essentially mimicking its methods.

```java
protected static final void debug(Logger logger, String message, Object... args) {
        if (isDebugEnabled) {
            logger.debug(getCallerMethodName() + "() " + message, args);
        }
 }
```

* I found this [PR SLF4J](https://github.com/qos-ch/slf4j/pull/29) which implements a `getLogger` that detects the calling Class. So at least that will probably be part of SLF4J in a future release. (One less thing we have to maintain on our own)

* The biggest concern that i have is the use of `getStackTrace()` which if i remember correctly is a pretty expensive call and could impact performance since it gets called twice every time something gets logged and even if a loglevel is disabled it gets called once in the current implementation. So if we want to merge this change we should at least reduce it to one `getStackTrace()` and only if the loglevel is enabled. Here is a good [stackoverflow](http://stackoverflow.com/questions/421280/how-do-i-find-the-caller-of-a-method-using-stacktrace-or-reflection) post on the topic also discussing different approaches that might be more performant.

Overall i like it but i'm don't think it's worth the performance trade off.

buraksarac:
@moio & @lucidd thanks for great comments.
*Stacktrace: Yes I have considered this. I was thinking few possible options:
 * javassist or similar(to much complexity to carry simplicity)
 * handling by preprocessor class loader so @Log annotation could be marked on methods or class (This also give opportunities using Context, since current design doesnt fit skipped )
 * Stacktrace: This was the easiest solution I end up. Instead every call I prefer to choose a cache to minise instance creation.

 **personal note: I think reflection is not a nightmare as before, most of libraries we are using they rely on reflection.

* Log level: You are right its quite laud at debug level by default but by my opinion if user of this dependency wants to say hey I want to see what com.suse.* package is doing then he set Trace level or related so he get more information. By default  it should be warn so you can tell to user 'Hey you gave me parameter X but there is a Y condition anyway I am going to process'.

*Formatter and other methods: If the class extends Logger it should get rid of missing methods or other futures.
 
 Since @lucidd  mentioned slf4j has roadmap and logging can be implemented maybe it is better to close this PR instead of implement changes and merge?

src/main/java/com/suse/saltstack/netapi/utils/LogUtils.java
renner: I agree, @moio: we don't need `@author` in comments, and a reference to the license we would put it only in some common copyright header that should be added to each source file I guess.

src/main/java/com/suse/saltstack/netapi/utils/LogUtils.java
```diff
@@ -0,0 +1,297 @@
...
+    /**
+     * Prints debug message by given logger<br/>
+     * Caller method name will be appended into starting of message
+     * 
+     * @param logger {@link Logger} instance
+     * @param message {@link String} message
+     * @param throwable {@link Throwable} throwable
+     */
+    protected static final void error(Logger logger, String message, Throwable throwable) {
+
```
renner: Is there a reason for putting an empty line at the beginning of each method?

buraksarac: It is because of my eclipse settings, I think I didnt have a proper configuration. I am waiting response if PR will be closed. If not I will commit required changes

src/main/java/com/suse/saltstack/netapi/utils/LogUtils.java
buraksarac: Thanks I will update comment template

src/main/java/com/suse/saltstack/netapi/utils/LogUtils.java
buraksarac: thaks I will update comment template

build.xml
buraksarac: I have added this line to include simplelogger.properties into classpath. I thought this should be the proper way using ant. As I am not expert of Any I was not able to show properties file to slf4j is there other way to this?

src/main/resources/simplelogger.properties
buraksarac: noted

PR #30, opened 2015-02-21
SaltStackClient:move generating JSON to explicit method

hustodemon:
...to slightly reduce duplicated code in SaltStackClient methods.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
@@ -12,10 +12,10 @@ import com.suse.saltstack.netapi.results.Result;
 import com.suse.saltstack.netapi.datatypes.Token;
 
 import com.google.gson.JsonArray;
-import com.google.gson.JsonObject;
-import com.google.gson.JsonPrimitive;
 
+import com.suse.saltstack.netapi.utils.ClientUtils;
 import java.net.URI;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
@@ -129,14 +129,18 @@ public class SaltStackClient {
      * @return authentication token as {@link Token}
      * @throws SaltStackException if anything goes wrong
      */
-    public Token login(String username, String password, String eauth)
+    public Token login(final String username, final String password, final String eauth)
             throws SaltStackException {
-        JsonObject json = new JsonObject();
-        json.addProperty("username", username);
-        json.addProperty("password", password);
-        json.addProperty("eauth", eauth);
+        Map<String, String> kwargs = new LinkedHashMap<String, String>() {
+            {
+                put("username", username);
+                put("password", password);
+                put("eauth", eauth);
+            }
+        };
         Result<List<Token>> result = connectionFactory
-                .create("/login", JsonParser.TOKEN, config).getResult(json.toString());
+                .create("/login", JsonParser.TOKEN, config)
+                .getResult(ClientUtils.makeJsonData(kwargs, null).toString());
 
         // For whatever reason they return a list of tokens here, take the first
         Token token = result.getResult().get(0);
@@ -234,32 +238,25 @@ public class SaltStackClient {
      * @return object representing the scheduled job
      * @throws SaltStackException if anything goes wrong
      */
-    public Job startCommand(String target, String function, List<String> args,
+    public Job startCommand(final String target, final String function, List<String> args,
             Map<String, String> kwargs) throws SaltStackException {
-        // Setup lowstate data to send as JSON
-        JsonObject json = new JsonObject();
-        json.addProperty("tgt", target);
-        json.addProperty("fun", function);
-        // Non-keyword arguments
-        if (args != null) {
-            JsonArray argsArray = new JsonArray();
-            for (String arg : args) {
-                argsArray.add(new JsonPrimitive(arg));
+        Map<String, String> allKwargs = new LinkedHashMap<String, String>() {
+            {
+                put("tgt", target);
+                put("fun", function);
             }
-            json.add("arg", argsArray);
-        }
-        // Keyword arguments
+        };
         if (kwargs != null) {
-            for (String key : kwargs.keySet()) {
-                json.addProperty(key, kwargs.get(key));
-            }
+            allKwargs.putAll(kwargs);
         }
+
         JsonArray jsonArray = new JsonArray();
-        jsonArray.add(json);
+        jsonArray.add(ClientUtils.makeJsonData(allKwargs, args));
 
         // Connect to the minions endpoint and send the above lowstate data
         Result<List<Job>> result = connectionFactory
-                .create("/minions", JsonParser.JOB,  config).getResult(jsonArray.toString());
+                .create("/minions", JsonParser.JOB,  config)
+                .getResult(jsonArray.toString());
 
         // They return a list of tokens here, we take the first
         return result.getResult().get(0);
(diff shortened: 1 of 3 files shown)


src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -97,14 +97,18 @@ public Token login(String username, String password)
...
      * @throws SaltStackException if anything goes wrong
      */
-    public Token login(String username, String password, String eauth)
+    public Token login(final String username, final String password, final String eauth)
             throws SaltStackException {
-        JsonObject json = new JsonObject();
-        json.addProperty("username", username);
-        json.addProperty("password", password);
-        json.addProperty("eauth", eauth);
+        Map<String, String> kwargs = new LinkedHashMap() {
```
moio: While I am a fan of this style of inline initialization of `Map`s and `List`s, I am not a big fan of `final` method parameters, [as pointed out in another PR](https://github.com/SUSE/saltstack-netapi-client-java/pull/28#commitcomment-9865167). Maybe it is just my personal taste, though. @renner, your idea?

hustodemon: Thanks for comment. Maybe I'm missing something, but I believe finalization of arguments doesn't leak outside of the method. It only means that arguments aren't overwriten inside the method. You can freely pass nonfinal fields to a method with final args. Isn't that correct?

renner: That's right @hustodemon, at least that was also my understanding of it. I would personally be fine with declaring parameters as `final` in order to ensure that these are not reassigned or changed inside that method. The code in #28 could even be simplified by that, or not? In there we explicitly create final versions of the input parameters for the async methods (`final String constUsername = username;`), which could be skipped if we declared them as final parameters.

hustodemon: Right, that's what I was commenting - making them final could prevent some boilerplate in #28.

The only issue with final arguments in this PR case is IMHO that it makes the method signatures even loooonger. If you think this is a problem, I'll revert back to non-final args (but will have to initialize the map the "traditional" way, i.e. without "double braces" syntax).

renner: I am fine with that, just wrap lines before reaching 92 characters and indent correctly. We might want to finalize the parameters of the (now merged) `*Async` methods in a separate commit or I will do it after we merged this.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -138,32 +142,25 @@ public Token login(String username, String password, String eauth)
...
-        // Setup lowstate data to send as JSON
-        JsonObject json = new JsonObject();
-        json.addProperty("tgt", target);
-        json.addProperty("fun", function);
-        // Non-keyword arguments
-        if (args != null) {
-            JsonArray argsArray = new JsonArray();
-            for (String arg : args) {
-                argsArray.add(new JsonPrimitive(arg));
+        Map<String, String> allKwargs = new LinkedHashMap() {
```
renner: Please add type parameters `<String, String>` to `new LinkedHashMap()` throughout this patch.

isbm: Diamond operator would work too, unless we are going to support EOF'ed Java 6.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
isbm: Personally I have no problems with the `final` keyword on the method signature and my personal taste to use it liberally. Some finds it too verbose, but others finds that the input parameters are never change, which is also good.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
renner: @isbm: The diamond operator cannot be used with anonymous classes, see e.g. [here] (http://stackoverflow.com/questions/9773733/double-brace-initialisation-anonymous-inner-class-with-diamond-operator).

renner:
Sorry, but could you please rebase this patch on top of the current master? I get some conflicts now when I pull the changes as #28 has been merged meanwhile. If you further fixed the missing type parameters I guess we can merge this. Thank you!

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
isbm: Bloody theory... :) Right, inference of a generic class's type arguments could influence the constraints on a generic constructor's type arguments. Well, then like you jus said: `<String, String>`.

hustodemon: Right, I initially wanted to use <> :)

hustodemon:
No problem!

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -309,36 +306,25 @@ public Job call() throws SaltStackException {
...
-        // Non-keyword arguments
-        if (args != null) {
-            JsonArray argsArray = new JsonArray();
-            for (String arg : args) {
-                argsArray.add(new JsonPrimitive(arg));
+    public Map<String,Object> run(final String username, final String password,
+            final String eauth, final String client, final String target,
+            final String function, List<String> args, Map<String, String> kwargs)
+            throws SaltStackException {
+        Map<String, String> allKwags = new LinkedHashMap<String, String>() {
```
renner: There is a typo here, it should probably be `allKwargs`?

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -385,4 +371,5 @@ public Job call() throws SaltStackException {
         };
         return executor.submit(callable);
     }
+
```
renner: Please, let's not add an empty line here.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
hustodemon: oh, no... lemme fix it.

src/test/java/com/suse/saltstack/netapi/utils/ClientUtilsTest.java
```diff
@@ -3,6 +3,14 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
+
```
renner: Please remove this line as well, let's put only one empty line to group imports.

src/test/java/com/suse/saltstack/netapi/utils/ClientUtilsTest.java
```diff
@@ -61,4 +68,37 @@ public void testStreamToString() {
...
+        List<String> args = new ArrayList<>();
+        args.add("foo");
+        args.add("bar");
+
+        JsonObject jsonObject = ClientUtils.makeJsonData(kwargs, args);
+
+        assertEquals(expected, jsonObject);
+    }
+
```
renner: ... and maybe this line as well? (sorry for nitpicking)

hustodemon: no problem, will fix (I'm used to do newlines so it doesn't burn my eyes).

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
moio: Alright, I'll update my mental style guide! Thanks for discussion.

renner:
I merged it, thanks again! Some type arguments were still missing in your test, added with 59974b1a252e596f544e25ed33c62293f2538ca8.

hustodemon:
Oh, I totally forgot. Thanks!

PR #31, opened 2015-02-22
Reorganize package structure

lucidd:
As suggested in #24 this PR reorganizes and renames some files as described in the commit messages.

If there are any suggestions / improvements please let me know :smile:

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/Constants.java
+++ b/src/main/java/com/suse/saltstack/netapi/Constants.java
@@ -1,9 +1,9 @@
-package com.suse.saltstack.netapi.client;
+package com.suse.saltstack.netapi;
 
 /**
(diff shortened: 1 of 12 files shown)


renner:
I am merging this patch, it is straight forward and brings (only) some refinements of the java packages structure. While I like to have well structured java packages and I think the suggested structure makes sense, we should maybe stop making it more granular at some point to avoid ending up with only one class per package :stuck_out_tongue_winking_eye:

lucidd:
@renner  yeah i noticed `JsonParser` and `Result` being pretty lonely now in their packages. I'm sure the parser package will have different parsers in the future to support the multiple formats the netapi provides i'm not so sure on the results package though. Maybe will will find a better package structure as the project grows.

PR #32, opened 2015-02-24
Add 'permissons' list ('perms' in json) to Token and extend the SaltStackClientTest with it

gborp:
Token datatype is extended with the 'permissions' field ('perms' in json). And also the SaltStackClientTest is extended to test it.


isbm:
Would be very nice to know what is this PR all about!

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
```diff
@@ -193,4 +196,4 @@ public void testRunResultAsync() throws SaltStackException {
         assertTrue(retvals.containsKey("minion-1"));
         assertEquals(retvals.get("minion-1"), true);
     }
-}
+}
```
isbm: No new line at the end of file.

src/main/java/com/suse/saltstack/netapi/datatypes/Token.java
```diff
@@ -31,4 +34,8 @@ public Date getStart() {
     public Date getExpire() {
         return expire;
     }
+
+    public List<String> getPerms() {
```
isbm: No documentation on the public method. I recon, you wanted to do something with permissions?

gborp: well, neither of the other methods do have a javadoc. I thought it's the convention for these kind of data holder classes.
about the "perms", I wasn't converted it into "permissons" as com.suse.saltstack.netapi.datatypes.Job also have "jid", not "jobID". So I thought it is the convention to let the parameters' name as they are in the json.

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
gborp: my bad. isn't there an official SUSE Java code formatter profile for Eclipse?

lucidd: @gborp i don't think there is an official profile but some contributors use Eclipse so maybe you can get a profile from them. There is also the style guide section in the readme which covers some formatting rules.

src/main/java/com/suse/saltstack/netapi/datatypes/Token.java
moio: @gborp: indeed I think you are right on both accounts. Naming-wise I would follow the "spirit" of the [API endpoint convention we talked about in PR 15](https://github.com/SUSE/saltstack-netapi-client-java/pull/15#issuecomment-74485966) - keep the original name unless there's a strong reason not to.

About missing Javadoc, we currently don't have it on any method on any class in the `datatype` package. The reason is that attributes are supposed to directly map to Salt's JSON (see above), thus any documentation is pure copy-and-paste exercise and it's not really needed. Even if at some point we decide to have it for completeness sake, I would suggest to add it to all classes in a separate PR, not on a case-by-case basis.

@isbm @renner: I propose to simply manually merge commit a8833da.

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
renner: @gborp Sorry, we currently don't distribute a profile, please just configure your IDE with the most important formatting rules that can be found in the [README] (https://github.com/SUSE/saltstack-netapi-client-java/blob/master/README.md).

src/main/java/com/suse/saltstack/netapi/datatypes/Token.java
renner: That's right, @moio, let's keep the original name. I will actually cherry-pick  4139a4a and a8833da then (= everything until the renaming).

renner:
I manually merged the first two commits as mentioned before, omitting the renaming of "perms" to "permissions". Thanks for this contribution!

PR #34, opened 2015-02-26
Switch project from ant to maven

lucidd:
This PR moves the project from ant to maven.

* Removes all the dependency jars since they are now managed by maven.
* Removes all Ant related files that are no longer needed.
* Adjusts `.gitignore` and `.travis.yml` for maven.

I also used this opportunity to move to the latest versions of the dependencies.

* `gson 2.3` -> `2.3.1`
* `httpclient 4.3.6` -> `4.4.0`

suggestions and feedback much appreciated :smile:


isbm:
Thank you for the refactoring, @lucidd. While it looks interesting, I would like myself to have an answers to the following questions:
1. What is the main advantage to move to Maven? I understand it is easier to find dependency and less hassle than with Ivy or you don't need to keep them somewhere around etc. While @moio's Tetra supposed to deal with the Maven packages, mere advantage is to easier development. But the impact here is still broader.
2. Can we also keep Ant and support it equally? It is still better to always have an option building package without Tetra magic, using just Ant and JPackage tools. There is a reason why GNU guys are building Java with plain Make, for example (I am not advocating this, just for your information). :-)

Thank you for the PR!

renner:
I like it, question is to @moio: Do you think we can drop build.xml completely yet? We will be able to build a package using maven, right? If we are not confident about that we should probably keep build.xml around.

@lucidd: How to generate an archive containing the sources, would you mind to add in the necessary `assembly` plugin?

dgiagio-druid:
@renner @moio Did you manage to build rpm packages with maven instead of ant? The last time we talked about this it wasn't possible because maven fetches libraries from the web.

dmacvicar:
We don't. Fedora has this problem solved but we don't (historical reasons, they bootstraped maven3 with maven2, and maven2 with ant over several years).

You can read more about it here: http://duncan.mac-vicar.com/2012/01/26/on-java-maven-jpp-and-rpm/

Our current plan, driven by @moio is tetra, to cut the dependency madness building from source, but using binary build-deps https://github.com/moio/tetra

lucidd:
@isbm 1. Like you said the dependency management is one advantage. As long as we are ok with the maven conventions and don't do anything too special it should result in a lot less configuration than we had with ant. 

2. We could keep the ant support but it would of course add extra work to keep them in sync. Also depending on what we use from maven it could potentially get hard to implement the same in ant. If we do keep both we should define what exactly it is we want/need from both systems and only keep that in sync.

@renner I will look into that and update the PR accordingly.

lucidd:
@renner I added a minimal config for `assembly` plugin that generates a source archive with `mvn assembly:single`. This can of course be configured more if the defaults don't fit our needs.

moio:
@lucidd, @dgiagio: let me clear this up.

I am sorry for unclear/conflicting messages regarding Maven use from our part so far.

Shipping software as Linux distro vendors is a bit more complex than just sharing a project on GitHub as you would do with any Java library - we want to give our users some extra guarantees with respect to maintainability and security. That places additional constraints to the source and binary distribution processes, such as being able to rebuild any piece of code without ever relying on binaries and without an Internet connection. It's a complex topic involving policy and technical issues and I don't want to cover it all here. Bottom line is that Maven usually means headaches, or at least more headaches than Ant, so it is typically not preferred by distro packagers.

This is true for most distros, with the notable exception of Arch, and is particularly bad on openSUSE and its commercial big brother SUSE Linux Enterprise. Needless to say, this is a pain point since Maven is very widespread.

I have been researching solutions for the last year or so, and I came up with one involving a major policy change and a minor helper tool named [tetra](https://github.com/moio/tetra). It is finally close to being fully blessed at least for SUSE Manager use but discussions are still ongoing - someone like @isbm rightfully wonders if an all-offline-by-default approach would be better, others like @dmacvicar and @renner rightfully wonder if tetra is mature enough, etc. Time will tell whether this really works or not.

As far as I am concerned I think it is viable right now for this project. I just created a package with the standard tetra workflow and it all went smoothly:

https://build.opensuse.org/package/show/home:SilvioMoioli:saltstack-api-java/saltstack-api-java

So from my POV this PR can be accepted, as soon as the current conflicts are solved (@lucidd can you please check and rebase if needed?).

Is it also OK for you guys?

isbm:
My **personal** opinion here is this: whilst packaging with Tetra works so far in this particular case, yet unless Tetra is actually blessed (and this is not happened so far), personally I would not be very happy ripping off Maven parts with pliers, reverting Ant in order to package it "the old way". For me is always more important to support package and put as less as possible complexity in a spec file, while development environment I always can build myself anyhow (i.e. I choose to trade-off at development time, rather then at support time). Therefore as long as we can really smoothly package this with Maven without putting extra work on this, I am _wholeheartedly_ support this PR.

So I would vote either Ant only or Ant + Maven. But Maven only.

dmacvicar:
@isbm Tetra is "blessed" for use in SUSE Manager.

I am fine that this is merged. Even if I found it unnecessary. It may increase the popularity of the library outside of the scope of our product.

However, maven means people will go crazy depending on the whole internet: http client libraries, containers, logging frameworks, etc. That we build it with tetra does not mean that we can add as many runtime dependencies as we want. Keep that in mind.

lucidd:
@moio sounds like an interesting problem. Now its more clear why this PR is so controversial. I think having at least a maven version around would be good so this project can help in testing tetra. Also like @dmacvicar  already mentioned it might help others to adopt this project. The question still remains if we want to have ant around a bit longer just in case. 
I also don't have a problem keeping the maven version in my own branch and updating it until we are sure what to do.

renner:
I am personally fine with merging this, as long as there is (more or less) feature equality with the previous build process. This is actually the case now with 7b5d6df it seems.

isbm:
Murphy says that `pom.xml` will be cool for us and `build.xml` will be hated by packagers. :-) I would at least keep Ant for a while. We always can drop it later.

moio:
@isbm, there's `git` for archiving stuff, I agree with @lucidd we should keep only minimal code around. Let's get rid of cruft, in case we really need to backpedal to Ant we will make use of `git revert`!

@dmacvicar adding new runtime dependencies is a whole different story and will be discussed in detail [as we did](https://github.com/SUSE/saltstack-netapi-client-java/pull/12#issuecomment-73852509) [in previous](https://github.com/SUSE/saltstack-netapi-client-java/pull/9#issuecomment-73056154) [PRs](https://github.com/SUSE/saltstack-netapi-client-java/pull/8#issuecomment-73234699), no worries.

By the way since we simply used `git add lib/some.jar` to add a runtime dependency with Ant, I don't expect Maven to lower this "barrier of entry" significantly.

Two developers OKed and one partly OKed, merging :-)

Thanks everyone

PR #38, opened 2015-02-26
add new test to SaltStackClientTest: testStartCommand()

gborp:
it tests SaltStackClient.startCommand(...).


src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
```diff
@@ -39,6 +40,8 @@
             "\"test.ping\"}]";
     final String JSON_RUN_RESPONSE = "{\"return\": [{\"minion-1\": true}]}";
 
+    final String JSON_START_COMMAND_RESPONSE = "{\"_links\":{\"jobs\":[{\"href\":\"/jobs/20150211105524392307\"}]},\"return\":[{\"jid\":\"20150211105524392307\",\"minions\":[\"myminion\"]}]}";
```
isbm: I would like to see it wrapped, please.

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
```diff
@@ -39,6 +40,8 @@
             "\"test.ping\"}]";
     final String JSON_RUN_RESPONSE = "{\"return\": [{\"minion-1\": true}]}";
```
isbm: Can we also remove this empty line?

PR #39, opened 2015-02-27
add new test to SaltStackClientTest: testStartCommand()

gborp:
(sorry for repeating the same PR as #38, I accidentally closed it)

it tests SaltStackClient.startCommand(...).


renner:
I merged it, thanks for your contribution! Note that an accidentally closed PR can normally just be reopened, there should be a button for that in the github web UI. Apart from that please make sure to always use *good* commit messages. In the above case of 33fbc4e I would clearly prefer to put the original message in the first line, and the recommit information below for further information. This makes it easier to understand what was done when viewing the history.

PR #41, opened 2015-03-02
new testStartCommandAsync() in SaltStackClientTest

gborp:
new testStartCommandAsync() in SaltStackClientTest. it tests SaltStackClient.startCommandAsync().

it's the async pair of the testStartCommand() in SaltStackClientTest, using the same input datas, doing the same checks.

### DIFF
--- a/src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
+++ b/src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
@@ -213,4 +213,24 @@ public class SaltStackClientTest {
         assertEquals(job.getJid(), "20150211105524392307");
         assertEquals(job.getMinions(), Arrays.asList("myminion"));
     }
+    
+    @Test
+    public void testStartCommandAsync() throws SaltStackException {
+    	 stubFor(post(urlEqualTo("/minions")).willReturn(
+                 aResponse().withStatus(HttpURLConnection.HTTP_OK)
+                         .withHeader("Content-Type", "application/json")
+                         .withBody(JSON_START_COMMAND_RESPONSE)));
+
+        Future<Job> future = client.startCommandAsync("*", "test.ping", null, null);
+        Job job;
+        try {
+        	job = future.get();
+        } catch (InterruptedException | ExecutionException e) {
+            throw new SaltStackException(e);
+        }
+
+        assertNotNull(job);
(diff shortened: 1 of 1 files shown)


src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
```diff
@@ -213,4 +213,24 @@ public void testStartCommand() throws SaltStackException {
...
+                 aResponse().withStatus(HttpURLConnection.HTTP_OK)
+                         .withHeader("Content-Type", "application/json")
+                         .withBody(JSON_START_COMMAND_RESPONSE)));
+
+        Future<Job> future = client.startCommandAsync("*", "test.ping", null, null);
+        Job job;
+        try {
+        	job = future.get();
+        } catch (InterruptedException | ExecutionException e) {
+            throw new SaltStackException(e);
```
moio: Why do you wrap `Exception`s here? I would just let the test fail on any `Exception` - wrapping them serves no purpose and adds one stack trace call to look at during debugging. Is there anything I am missing?

gborp: I've tried to follow the unwritten conventions of the SaltStackClientTest class, the other test*Async methods do the same catch-wrap-throw thing, possibly to reduce the number of thrown exception of the test method. (As it is a unit-test method, I don't think it makes sense.)
it's a styling question, I can't judge it.
of course it can be removed and add these 2 exception types to the methods' "throws" list.

please someone tell me which one is preferred:
A, less, simplier code line (without the catch-wrap-throw) (should I change the other test*Async methods too? or should it be a separate PR?)
B, follow the style of the other *Async methods.

I would choose A, with or without modifying the other test*Async methods.

moio: Sorry @gborp, I didn't notice other methods were doing the same. I would say, let's merge this and then have another PR where we remove all needless `Exception` wrapping (option A).

In my opinion, tests should throw `Exception` - listing specific subclasses is just maintenance burden.

moio:
Merged manually due to a trivial conflict with another PR.

Please note that I added commit 0564fce on top because you still had tabs in your file, make sure you always indent with 4 spaces from now on.

Thanks!

Issue #43, opened 2015-03-04
Issues with tests


lucidd:
While working on #42 i found a few problems that i want to discuss and get some feedback first before changing anything. 

#### `testLoginFailure`

I'm not completely sure what [testLoginFailure](https://github.com/SUSE/saltstack-netapi-client-java/blob/master/src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java#L72-L80) tests. From what i can tell i would say it was supposed to test logging in with wrong credentials but since it throws the exception for all return codes other than `OK` and `ACCEPTED` this test passes in a lot of cases when it shouldn't.

Of course we could check if the exception message contains the right error code but i think it would be better to not leak the HTTP return codes and instead throw more meaningful exceptions so it is easier to handle different errors differently.

#### `stubFor` vs `verify`

Wiremock allows to specify matching rules in both `stubFor` and `verify`. While `stubFor`
is basically a router to determine which requests get served what response, verify serves the role of an assertion for the test. The problem is when a request does not match the `stubFor` it gets served a `404` which results in the test failing with a `SaltStackException` containing the `404`. There is not a lot of information about what went wrong. On the other hand if a request does not match the `verify` rules both expected and actual request(s) are printed as part of the failure.

So my suggestion would be using a `stubFor` that accepts all requests and shifting the the verification completely to verify.

moio:
Your points make sense to me. Feel free to proceed fixing them (in case they are still relevant), otherwise let's simply keep this issue open not to forget about it.

Thanks

lucidd:
It should be still relevant. I will prepare a PR as soon as i have some time.

lucidd:
Fixed by #66

PR #44, opened 2015-03-05
/stats endpoint implementation

lucidd:
This PR implements access to the netapi `/stats` endpoint. I couldn't find any documentation on the return type of this call so all the datatypes are "best guess" implementations based on the `json` returned by my local saltmaster. I also left the `Request` field of `cherrypy.Applications` unimplemented and added a todo since the `json` only contains an empty object so i don't know it's structure.

#### Aditional thoughts / Improvements

* `getBindAddress` could return `InetSocketAddress` instead of a `String`
* Exposing `getThreads` feels redundant since `getWorkerThreads().size()` should give the same information.
* `Uptime` could be represented with `java.time.Duration` when we move to Java 8.
* Redundant information like in `threads` /  `workerThreads` and `uptime` / `startTime` / `currentTime` could be checked against each other for validation to filter some "bad" responses.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
@@ -5,6 +5,7 @@ import com.suse.saltstack.netapi.client.impl.HttpClientConnectionFactory;
 import com.suse.saltstack.netapi.config.ClientConfig;
 import static com.suse.saltstack.netapi.config.ClientConfig.*;
 import com.suse.saltstack.netapi.config.ProxySettings;
+import com.suse.saltstack.netapi.datatypes.cherrypy.Stats;
 import com.suse.saltstack.netapi.exception.SaltStackException;
 import com.suse.saltstack.netapi.parser.JsonParser;
 import com.suse.saltstack.netapi.datatypes.Job;
@@ -385,4 +386,33 @@ public class SaltStackClient {
         };
         return executor.submit(callable);
     }
+
+  /**
+  * Query statistics from the CherryPy Server.
+  *
+  * GET /stats
+  *
+  * @return The {@link Stats} object.
+  * @throws SaltStackException if anything goes wrong
+  */
+  public Stats stats() throws SaltStackException {
+      return connectionFactory.create("/stats", JsonParser.STATS, config).getResult();
+  }
+
+  /**
+  * Asynchronously query statistics from the CherryPy Server.
+  *
+  * GET /stats
+  *
+  * @return Future containing the {@link Stats} object.
+  */
+  public Future<Stats> statsAsync() {
+      Callable<Stats> callable = new Callable<Stats>() {
+          @Override
+          public Stats call() throws SaltStackException {
+              return stats();
+          }
+      };
+      return executor.submit(callable);
+  }
 }
--- a/src/main/java/com/suse/saltstack/netapi/datatypes/cherrypy/Applications.java
+++ b/src/main/java/com/suse/saltstack/netapi/datatypes/cherrypy/Applications.java
@@ -1,0 +1,121 @@
+package com.suse.saltstack.netapi.datatypes.cherrypy;
+
+import com.google.gson.annotations.SerializedName;
+
+import java.util.Date;
+import java.util.Map;
+
+public class Applications {
+
+    @SerializedName("Uptime")
+    private double uptime;
+
+    @SerializedName("Bytes Read/Second")
+    private double readsPerSecond;
+
+    @SerializedName("Current Time")
+    private Date currentTime;
+
+    @SerializedName("Server Version")
+    private String serverVersion;
+
+    @SerializedName("Total Time")
+    private double totalTime;
+
+    @SerializedName("Enabled")
+    private boolean enabled;
+
+    @SerializedName("Start Time")
+    private Date startTime;
+
+    @SerializedName("Bytes Written/Second")
+    private double writesPerSecond;
+
+    @SerializedName("Total Bytes Read")
+    private int totalBytesRead;
+
+    @SerializedName("Current Requests")
+    private int currentRequests;
+
+    @SerializedName("Total Requests")
+    private int totalRequests;
+
+    @SerializedName("Requests")
+    Map<String, Request> requests;
+
+    @SerializedName("Bytes Read/Request")
+    private double readsPerRequest;
+
+    @SerializedName("Total Bytes Written")
+    private int totalBytesWritten;
+
+    @SerializedName("Requests/Second")
+    private double requestsPerSecond;
+
+    @SerializedName("Bytes Written/Request")
+    private double writesPerRequest;
+
+    public double getUptime() {
+        return uptime;
+    }
+
+    public double getReadsPerSecond() {
+        return readsPerSecond;
+    }
+
+    public Date getCurrentTime() {
+        return currentTime;
+    }
+
+    public String getServerVersion() {
+        return serverVersion;
+    }
+
+    public double getTotalTime() {
+        return totalTime;
+    }
+
+    public boolean isEnabled() {
+        return enabled;
+    }
+
+    public Date getStartTime() {
+        return startTime;
+    }
+
+    public double getWritesPerSecond() {
+        return writesPerSecond;
+    }
+
+    public int getTotalBytesRead() {
+        return totalBytesRead;
+    }
+
+    public int getCurrentRequests() {
+        return currentRequests;
+    }
+
+    public int getTotalRequests() {
+        return totalRequests;
+    }
+
+    public Map<String, Request> getRequests() {
+        return requests;
+    }
+
+    public double getReadsPerRequest() {
+        return readsPerRequest;
+    }
+
+    public int getTotalBytesWritten() {
+        return totalBytesWritten;
+    }
+
+    public double getRequestsPerSecond() {
+        return requestsPerSecond;
+    }
+
+    public double getWritesPerRequest() {
+        return writesPerRequest;
+    }
+}
--- a/src/main/java/com/suse/saltstack/netapi/datatypes/cherrypy/Request.java
+++ b/src/main/java/com/suse/saltstack/netapi/datatypes/cherrypy/Request.java
@@ -1,0 +1,64 @@
+package com.suse.saltstack.netapi.datatypes.cherrypy;
+
+import com.google.gson.annotations.SerializedName;
+
+import java.util.Date;
+
+public class Request {
+
+    @SerializedName("Bytes Read")
+    private Integer bytesRead;
+
+    @SerializedName("Bytes Written")
+    private Integer bytesWritten;
+
+    @SerializedName("Response Status")
+    private String responeStatus;
+
+    @SerializedName("Start Time")
+    private Date startTime;
+
+    @SerializedName("End Time")
+    private Date endTime;
+
+    @SerializedName("Client")
+    private String client;
+
+    @SerializedName("Processing Time")
+    private double processingTime;
+
+    @SerializedName("Request-Line")
+    private String requestLine;
+
+    public Integer getBytesRead() {
+        return bytesRead;
+    }
+
+    public Integer getBytesWritten() {
+        return bytesWritten;
+    }
+
+    public String getResponeStatus() {
+        return responeStatus;
+    }
+
+    public Date getStartTime() {
+        return startTime;
+    }
+
+    public Date getEndTime() {
+        return endTime;
+    }
+
+    public String getClient() {
+        return client;
+    }
+
+    public double getProcessingTime() {
+        return processingTime;
+    }
+
+    public String getRequestLine() {
+        return requestLine;
+    }
+}
--- a/src/main/java/com/suse/saltstack/netapi/datatypes/cherrypy/Stats.java
+++ b/src/main/java/com/suse/saltstack/netapi/datatypes/cherrypy/Stats.java
@@ -1,0 +1,24 @@
+package com.suse.saltstack.netapi.datatypes.cherrypy;
+
+public class Stats {
+
+    private final Applications applications;
+    private final HttpServer httpServer;
+
+    public Stats(Applications applications, HttpServer httpServer)
+            throws IllegalArgumentException {
+        if (applications == null || httpServer == null) {
+            throw new IllegalArgumentException("applications and httpServer must not be null");
+        }
+        this.applications = applications;
+        this.httpServer = httpServer;
+    }
+
+    public Applications getApplications() {
+        return applications;
+    }
+
+    public HttpServer getHttpServer() {
+        return httpServer;
+    }
+}
--- a/src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
+++ b/src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
@@ -5,10 +5,14 @@ import com.google.gson.GsonBuilder;
 import com.google.gson.JsonDeserializationContext;
 import com.google.gson.JsonDeserializer;
 import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
 import com.google.gson.JsonParseException;
 import com.google.gson.reflect.TypeToken;
 import com.suse.saltstack.netapi.datatypes.Job;
 import com.suse.saltstack.netapi.datatypes.Token;
+import com.suse.saltstack.netapi.datatypes.cherrypy.Stats;
+import com.suse.saltstack.netapi.datatypes.cherrypy.Applications;
+import com.suse.saltstack.netapi.datatypes.cherrypy.HttpServer;
 import com.suse.saltstack.netapi.results.*;
 
 import java.io.BufferedReader;
@@ -36,6 +40,8 @@ public class JsonParser<T> {
         new JsonParser<>(new TypeToken<Result<List<Job>>>(){});
     public static final JsonParser<Result<List<Map<String,Object>>>> RETVALS =
         new JsonParser<>(new TypeToken<Result<List<Map<String,Object>>>>(){});
+    public static final JsonParser<Stats> STATS =
+        new JsonParser<>(new TypeToken<Stats>(){});
 
     private final TypeToken<T> type;
     private final Gson gson;
@@ -48,7 +54,9 @@ public class JsonParser<T> {
     public JsonParser(TypeToken<T> type){
         this.type = type;
         this.gson = new GsonBuilder()
-                .registerTypeAdapter(Date.class, new SaltStackDateDeserializer()).create();
+                .registerTypeAdapter(Date.class, new SaltStackDateDeserializer())
+                .registerTypeAdapter(Stats.class, new StatsDeserializer())
+                .create();
     }
 
     /**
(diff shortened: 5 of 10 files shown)


src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -385,4 +386,19 @@ public Job call() throws SaltStackException {
         };
         return executor.submit(callable);
     }
+
+  public Stats stats() throws SaltStackException {
+      return connectionFactory.create("/stats", JsonParser.STATS, config).getResult();
+  }
+
+  public Future<Stats> statsAsync() {
```
isbm: Javadoc?

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -385,4 +386,19 @@ public Job call() throws SaltStackException {
         };
         return executor.submit(callable);
     }
+
+  public Stats stats() throws SaltStackException {
```
isbm: I am wondering if there is needed synchronous `stats()` at all? Maybe `statsAsync()` can be `stats()` and `stats()` can be removed completely? What is your opinion here?

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -385,4 +386,19 @@ public Job call() throws SaltStackException {
...
+  public Future<Stats> statsAsync() {
+      Callable<Stats> callable = new Callable<Stats>() {
+          @Override
+          public Stats call() throws SaltStackException {
+              return stats();
+          }
+      };
+      return executor.submit(callable);
+  }
+
```
isbm: Sometimes you add this empty line, sometimes not. It would be better to stay consistent with the rest of the project. Sorry for the nitpick. :)

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -81,4 +89,29 @@ public Date deserialize(JsonElement jsonElement, Type type,
...
+        public Stats deserialize(JsonElement jsonElement, Type type,
+                JsonDeserializationContext jsonDeserializationContext)
+                throws JsonParseException {
+            try {
+                JsonObject stats = jsonElement.getAsJsonObject();
+                Applications app = gson.fromJson(stats.get("CherryPy Applications"), Applications.class);
+                HttpServer server = null;
+                for(Map.Entry<String, JsonElement> entry : stats.entrySet()){
+                    String key = entry.getKey();
+                    if(key.startsWith("CherryPy HTTPServer ")){
```
isbm: Looks like a static final constant checking. And above, line 101 too.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -81,4 +89,29 @@ public Date deserialize(JsonElement jsonElement, Type type,
...
+                        server = gson.fromJson(entry.getValue(), HttpServer.class);
+                        break;
+                    }
+                }
+                return new Stats(app, server);
```
isbm: Just in case you missed it, here is a risk to return `Stats(app, null)` instead.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
lucidd: Totally forgot that thanks for the reminder. :)

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
lucidd: So far all the methods have a sync and async implementation so i wouldn't make an exception here to keep the consistency.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
lucidd: No problem somebody has to point it out :) Will fix that. `ParsingException` and `ClientConfig` also have this empty line but i will submit another PR for those.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
lucidd: I actually had them as `static final` at first but when i realized that i just used them once i thought inlining them would be better. Is there still an advantage of making them `static final` even if they are only used once?

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
lucidd: I didn't miss that and while i'm not particularly happy with it i don't think there is a better solution until we can use Java 8s `Optional`. But i'm open for suggestions :)

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
moio: I agree with @lucidd here. @isbm why do you think a synchronous method is superfluous?

src/main/java/com/suse/saltstack/netapi/datatypes/cherrypy/Applications.java
```diff
@@ -0,0 +1,117 @@
...
+
+    @SerializedName("Current Requests")
+    private int currentRequests;
+
+    @SerializedName("Total Requests")
+    private int totalRequests;
+
+    //TODO: implement requests
+    //@SerializedName("Requests")
+    //private Object requests;
```
moio: Maybe it's just a personal habit but I really dislike TODO comments and, even more, committed but commented out code.

For TODOs we have GitHub issues, which are harder to forget about and easier for others to pick up. Project management does not work well in source code.

For disabled code we have git - there's no need to keep unused, untested an probably not working code around.

So please remove these lines or consider handling Requests properly.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
moio: I also personally tend to prefer inlining and refactoring out as constant as soon as I add a second use, but no strong opinion on this one. I would accept either way. @renner?

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
moio: *[I am being a bit more verbose here since similar discussions happen routinely, so I hope to set this topic straight once and for all]*

Deciding whether to return `null` from an API should always be driven by the following question, which I also encourage to apply here:

**Is `null` expected as a particular but possible case or does it mean that an error has happened somewhere?**

If `null` is just one output possibility and really indicates a missing value that is optional in its nature, but it signals no error per se, then you should return `null`, document it and have the caller deal with it. If caller's code ends up with a nasty `NullPointerException` because they forgot to handle the `null` case properly, that's their fault, they should fix it and you don't have to worry. One day Java 8 arrives and you will be able to do the same, with compiler assistance by using `Optional`.

If `null` in your output is not really reasonable, but rather tells that your code has encountered a problem, then you should not return it. As usual you have two possibilities: either handle the error in your method, if you have enough information to do so, or signal the problem to the caller by throwing an appropriate `Exception` and move the error handling burden one layer up.

@lucidd: in this particular case, can you tell if a server element is always expected to be returned from the CherryPy API? If it is always expected, then you should check that it is really there and throw `Exception` if it is not - CherryPy is returning incomplete data so there's a problem below your level, and whoever calls you should know as you can't possibly handle the issue in here.

Otherwise you should be able to answer this question: in what reasonable cases do we expect a server to be missing from CherryPy's API output?

moio:
@lucidd hey thanks for this contribution! I am happy to see some functional extensions to the library after many cleanups and refactorings, :+1:!

Please fix issues discussed in line comments as usual. Apart from those points, which are minor anyway, my most important comment is:

> I couldn't find any documentation on the return type of this call so all the datatypes are "best guess" implementations based on the `json` returned by my local saltmaster. I also left the `Request` field of `cherrypy.Applications` unimplemented and added a todo since the `json` only contains an empty object so i don't know it's structure.

Use the Source, Luke!

First, you should check CherryPy's source code to make sure your assumptions are correct. Being open source means we don't have to treat it as a black box!

Second, if you identified a documentation lack either in CherryPy or in rest_cherrypy, you should really take the chance to contribute your comments to those projects - opening bugs/issues, contacting people via IRC/mailing list, whatever is appropriate to hep them improve.

And maybe tell them "Hey, you know what? There's a cool new Java API wrapper here"!

Happy hacking.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
renner: I would prefer to convert those to constants, even if they are used only once. There is no real advantage of that, it's just a personal preference, I like to keep those strings together in one place. Further in case one of them needs to be adapted slightly, the code can stay untouched.

src/main/java/com/suse/saltstack/netapi/datatypes/cherrypy/Applications.java
renner: For me it is OK to have some TODO comments in place in the code, wherever it makes sense for the reader to know that there is something missing for instance. I very much agree with @moio though about commented out code: it should not be committed if it is commented out.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
renner: I also agree with @lucidd and @moio as we currently have two versions of every method.

isbm: I think @lucidd argued it right. No more questions here. :)

src/main/java/com/suse/saltstack/netapi/datatypes/cherrypy/Applications.java
isbm: I was actually about to duplicate @moio but then I refrained (knowing he will do that anyway :-) ). Personally I am totally fine to have todos marking in the code (after all IDEs are marking them as tasks), as long as there is a new PR coming right away soon. It is indeed easier and much more comfortable to double-click in IDE a task, jumping to a TODO and implement it, rather than forget some issue in a GitHub through the hassle. But yes, I agree with @moio  about other comments. Better would be to move all the stuff to a GitHub issues or Gist etc, like:
`//TODO: Implement scary Godzilla, https://github.com/SUSE/saltstack-netapi-client-java/issues/666`

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
lucidd: @moio You are right that's a really good way of putting it. Since it says CherryPy HTTPServer i would guess thats is statistics about the http server we are using to make the api calls. So unless those statistics can be disabled in some config i would expect them always to be there if we get data back from the server. I will change it to throw an exception

src/main/java/com/suse/saltstack/netapi/datatypes/cherrypy/Applications.java
lucidd: @moio Good point i think its a good idea to utilize `issues` more so far there is only one ( #43 shameless plug) by me which has not seen any discussion yet.
I mostly based that on how the Dates in `Token` got handled which was similarly commented out marked with a todo so i thought it would it would be ok but now i know better :). 

I really like the idea of @isbm adding a `TODO` mark pointing to the issue. Its mostly personal preference tho since i like to use the todolist of IDEs or grepping for todos.

I will remove it completely but i still would like your thoughts on @isbm  suggestion.  :)

moio: Fair enough, leave the `TODO`, optionally add an issue, remove dead code.

lucidd:
@moio Haha funny but you are right again i should really do that. I will look into it more as soon as i addressed the other comments.

lucidd:
Ok i addressed all the problems and opened #47  for tracking progress for implementing the missing `Requests` field.

lucidd:
Ok after some digging in the [source](https://github.com/cherrypy/cherrypy/blob/ec3a098e84277e8791734a562ddf7c0bf1757440/cherrypy/lib/cpstats.py#L325-L335) i found the structure of the `Requests`. I also found out that stats collecting is disabled by default and you have to enable it (as documented [here](http://docs.saltstack.com/en/latest/ref/netapi/all/salt.netapi.rest_cherrypy.html#a-rest-api-for-salt)). So now `Requests` are implemented and the tests are updated with some real data returned from `/stats`.

This closes #47

src/main/java/com/suse/saltstack/netapi/datatypes/cherrypy/Request.java
```diff
@@ -0,0 +1,65 @@
+package com.suse.saltstack.netapi.datatypes.cherrypy;
+
+
```
renner: Maybe remove one of those empty lines here.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -81,4 +89,39 @@ public Date deserialize(JsonElement jsonElement, Type type,
...
+                    return new Stats(app, server);
+                } else {
+                    throw new JsonParseException("CherryPy HTTPServer is null");
+                }
+            } catch (IllegalStateException e) {
+                throw new JsonParseException(e);
+            }
+        }
+    }
+
```
renner: This empty line could also be removed.

renner:
The patch looks very good to me, thanks @lucidd! I didn't try it out yet, but I guess we can merge it. Consider to remove those empty lines before though.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -81,4 +89,39 @@ public Date deserialize(JsonElement jsonElement, Type type,
...
+                    String key = entry.getKey();
+                    if(key.startsWith(CP_SERVER_PREFIX)){
+                        server = gson.fromJson(entry.getValue(), HttpServer.class);
+                        break;
+                    }
+                }
+                if(server != null) {
+                    return new Stats(app, server);
+                } else {
+                    throw new JsonParseException("CherryPy HTTPServer is null");
```
isbm: Well, I would say such message does not helps other than just for a low-level debugging purposes. It rather should say that the HTTP server is not available and possibly hint why this particular exception happened (e.g. SSL wrong, connection timed out, no socket connection etc).

src/main/java/com/suse/saltstack/netapi/datatypes/cherrypy/Stats.java
```diff
@@ -0,0 +1,20 @@
+package com.suse.saltstack.netapi.datatypes.cherrypy;
+
+public class Stats {
+
+    private final Applications applications;
+    private final HttpServer httpServer;
+
+    public Stats(Applications applications, HttpServer httpServer) {
+        this.applications = applications;
+        this.httpServer = httpServer;
```
isbm: Looks like here it should check if apps and server are not null, throw the exception otherwise. This will save you extra code everywhere to check if the objects are there as well as not to initiate `Stats` class with `null`, like you do below in `StatsDeserializer`.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -81,4 +89,39 @@ public Date deserialize(JsonElement jsonElement, Type type,
...
+                HttpServer server = null;
+                for(Map.Entry<String, JsonElement> entry : stats.entrySet()){
+                    String key = entry.getKey();
+                    if(key.startsWith(CP_SERVER_PREFIX)){
+                        server = gson.fromJson(entry.getValue(), HttpServer.class);
+                        break;
+                    }
+                }
+                if(server != null) {
+                    return new Stats(app, server);
```
isbm: And if `app` is null?

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -81,4 +89,39 @@ public Date deserialize(JsonElement jsonElement, Type type,
...
+                throws JsonParseException {
+            try {
+                JsonObject stats = jsonElement.getAsJsonObject();
+                Applications app = gson.fromJson(stats.get(CP_APPLICATIONS), Applications.class);
+                HttpServer server = null;
+                for(Map.Entry<String, JsonElement> entry : stats.entrySet()){
+                    String key = entry.getKey();
+                    if(key.startsWith(CP_SERVER_PREFIX)){
+                        server = gson.fromJson(entry.getValue(), HttpServer.class);
+                        break;
```
isbm: Here I would just return `new Stats(...);` with added `null`-check, as suggested above in `Stats` constructor instead. That would minimize your code as well as it would eliminate checking if `Stats` container class returns non-null objects for further access.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -81,4 +89,39 @@ public Date deserialize(JsonElement jsonElement, Type type,
...
+        public Stats deserialize(JsonElement jsonElement, Type type,
+                JsonDeserializationContext jsonDeserializationContext)
+                throws JsonParseException {
+            try {
+                JsonObject stats = jsonElement.getAsJsonObject();
+                Applications app = gson.fromJson(stats.get(CP_APPLICATIONS), Applications.class);
+                HttpServer server = null;
+                for(Map.Entry<String, JsonElement> entry : stats.entrySet()){
+                    String key = entry.getKey();
+                    if(key.startsWith(CP_SERVER_PREFIX)){
```
renner: Let's put blanks after `for` and `if` as well as before the `{` (throughout the whole patch).

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
lucidd: I think `HTTP server is not available` gives the wrong impression because obviously it just returned me some `json` so the  HttpServer must be up and running. I don't think a valid response would ever have the server missing.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
lucidd: Right app being `null` should also be an error since its always present even when stats collecting is disabled. I will adapt your suggesting of moving the `null` checks into the constructor that way we can be sure that app and server is never `null` if we have a stats object. :)

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
isbm: But so is `CherryPy HTTPServer is null` — how `null` can return JSON at first place? :) I would suggest to define that the object in the data wasn't defined or something that direction, for example.

lucidd:
* Formatted all `for` and `if`s introduced by this patch.
* Removed unnecessary empty lines.
* Moved the error handling into the `Stats` constructor.

Thanks all for the feedback. If there is anything else let me know.

renner:
I merged it, thanks for this patch! Personally I disagree with @isbm on the error handling topic and would have preferred to keep the null checks in the deserializer, throwing an exception from there in case of an error with the Stats class being just a container. We might discuss that again and might even change it back later on, but this is only a minor detail so I am merging the patch now as it is.

PR #45, opened 2015-03-05
Add support for querrying jobs

hustodemon:
...by parsing responses to GET request on '/jobs' resource.

- The 'Job' datatype was renamed to more meaningful 'JobMinions'
  datatype.
- New datatype 'Job' was created into which the '/jobs' resource is
  parsed.
- Added corresponding parser and parser test.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
@@ -8,7 +8,7 @@ import com.suse.saltstack.netapi.datatypes.Keys;
 import com.suse.saltstack.netapi.datatypes.cherrypy.Stats;
 import com.suse.saltstack.netapi.exception.SaltStackException;
 import com.suse.saltstack.netapi.parser.JsonParser;
-import com.suse.saltstack.netapi.datatypes.Job;
+import com.suse.saltstack.netapi.datatypes.JobMinions;
 import com.suse.saltstack.netapi.results.Result;
 import com.suse.saltstack.netapi.datatypes.Token;
 
@@ -192,7 +192,8 @@ public class SaltStackClient {
     }
 
     /**
-     * Generic interface to start any execution command and immediately return the job id.
+     * Generic interface to start any execution command and immediately return an object
+     * representing the scheduled job.
      *
      * POST /minions
      *
@@ -203,8 +204,8 @@ public class SaltStackClient {
      * @return object representing the scheduled job
      * @throws SaltStackException if anything goes wrong
      */
-    public Job startCommand(final String target, final String function, List<String> args,
-            Map<String, String> kwargs) throws SaltStackException {
+    public JobMinions startCommand(final String target, final String function,
+            List<String> args, Map<String, String> kwargs) throws SaltStackException {
         Map<String, String> props = new LinkedHashMap<String, String>() {
             {
                 put("tgt", target);
@@ -216,8 +217,8 @@ public class SaltStackClient {
         jsonArray.add(ClientUtils.makeJsonData(props, kwargs, args));
 
         // Connect to the minions endpoint and send the above lowstate data
-        Result<List<Job>> result = connectionFactory
-                .create("/minions", JsonParser.JOB,  config)
+        Result<List<JobMinions>> result = connectionFactory
+                .create("/minions", JsonParser.JOB_MINIONS,  config)
                 .getResult(jsonArray.toString());
 
         // They return a list of tokens here, we take the first
@@ -225,7 +226,8 @@ public class SaltStackClient {
     }
 
     /**
-     * Asynchronously start any execution command and immediately return the job id
+     * Asynchronously start any execution command and immediately return an object
+     * representing the scheduled job.
      *
      * POST /minions
      *
@@ -233,13 +235,13 @@ public class SaltStackClient {
      * @param function the function to execute
      * @param args list of non-keyword arguments
      * @param kwargs map containing keyword arguments
-     * @return Future containing the scheduled job {@link Job}
+     * @return Future containing the scheduled job {@link JobMinions}
      */
-    public Future<Job> startCommandAsync(final String target, final String function,
+    public Future<JobMinions> startCommandAsync(final String target, final String function,
             final List<String> args, final Map<String, String> kwargs) {
-        Callable<Job> callable = new Callable<Job>() {
+        Callable<JobMinions> callable = new Callable<JobMinions>() {
             @Override
-            public Job call() throws SaltStackException {
+            public JobMinions call() throws SaltStackException {
                 return startCommand(target, function, args, kwargs);
             }
         };
@@ -251,11 +253,12 @@ public class SaltStackClient {
      *
      * GET /job/<job-id>
      *
-     * @param job {@link Job} object representing scheduled job
+     * @param job {@link JobMinions} object representing scheduled job
      * @return Map key: minion id, value: command result from that minion
      * @throws SaltStackException if anything goes wrong
      */
-    public Map<String, Object> getJobResult(final Job job) throws SaltStackException {
+    public Map<String, Object> getJobResult(final JobMinions job)
+            throws SaltStackException {
         return getJobResult(job.getJid());
     }
 
--- a/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
@@ -8,6 +8,7 @@ import com.suse.saltstack.netapi.datatypes.Keys;
 import com.suse.saltstack.netapi.datatypes.cherrypy.Stats;
 import com.suse.saltstack.netapi.exception.SaltStackException;
 import com.suse.saltstack.netapi.parser.JsonParser;
+import com.suse.saltstack.netapi.datatypes.Job;
 import com.suse.saltstack.netapi.datatypes.JobMinions;
 import com.suse.saltstack.netapi.results.Result;
 import com.suse.saltstack.netapi.datatypes.Token;
@@ -278,6 +279,32 @@ public class SaltStackClient {
 
         // A list with one element is returned, we take the first
         return result.getResult().get(0);
+    }
+
+    /**
+     * Get previously run jobs.
+     * @return map containing run jobs keyed by job id.
+     * @throws SaltStackException if anything goes wrong
+     */
+    public Map<String, Job> getJobs() throws SaltStackException {
+        Result<List<Map<String, Job>>> result = connectionFactory
+                .create("/jobs", JsonParser.JOBS, config)
+                .getResult();
+        return result.getResult().get(0);
+    }
+
+    /**
+     * Get previously run jobs.
+     * @return future with a map containing run jobs keyed by job id.
+     */
+    public Future<Map<String, Job>> getJobsAsync() {
+        Callable<Map<String, Job>> callable = new Callable<Map<String, Job>>() {
+            @Override
+            public Map<String, Job> call() throws SaltStackException {
+                return getJobs();
+            }
+        };
+        return executor.submit(callable);
     }
 
     /**
--- a/src/main/java/com/suse/saltstack/netapi/datatypes/Job.java
+++ b/src/main/java/com/suse/saltstack/netapi/datatypes/Job.java
@@ -1,0 +1,47 @@
+package com.suse.saltstack.netapi.datatypes;
+
+import com.google.gson.annotations.SerializedName;
+
+// TODO - handle dates too (they are represented in really
+// exotic format: (2015, Mar 04 19:28:29.724698))
+
+/**
+ * Representation of a previously run job.
+ */
+public class Job {
+
+    @SerializedName("Function")
+    private String function;
+
+    @SerializedName("Target")
+    private String target;
+
+    @SerializedName("Target-type")
+    private String targetType;
+
+    @SerializedName("User")
+    private String user;
+
+    @SerializedName("Arguments")
+    private Arguments arguments;
+
+    public String getFunction() {
+        return function;
+    }
+
+    public String getTarget() {
+        return target;
+    }
+
+    public String getTargetType() {
+        return targetType;
+    }
+
+    public String getUser() {
+        return user;
+    }
+
+    public Arguments getArguments() {
+        return arguments;
+    }
+}
--- a/src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
+++ b/src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
@@ -8,7 +8,7 @@ import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParseException;
 import com.google.gson.reflect.TypeToken;
-import com.suse.saltstack.netapi.datatypes.Job;
+import com.suse.saltstack.netapi.datatypes.JobMinions;
 import com.suse.saltstack.netapi.datatypes.Keys;
 import com.suse.saltstack.netapi.datatypes.Token;
 import com.suse.saltstack.netapi.datatypes.cherrypy.Applications;
@@ -36,8 +36,8 @@ public class JsonParser<T> {
             new JsonParser<>(new TypeToken<Result<String>>(){});
     public static final JsonParser<Result<List<Token>>> TOKEN =
             new JsonParser<>(new TypeToken<Result<List<Token>>>(){});
-    public static final JsonParser<Result<List<Job>>> JOB =
-            new JsonParser<>(new TypeToken<Result<List<Job>>>(){});
+    public static final JsonParser<Result<List<JobMinions>>> JOB_MINIONS =
+            new JsonParser<>(new TypeToken<Result<List<JobMinions>>>(){});
     public static final JsonParser<Result<List<Map<String, Object>>>> RETVALS =
             new JsonParser<>(new TypeToken<Result<List<Map<String, Object>>>>(){});
     public static final JsonParser<Stats> STATS =
--- a/src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
+++ b/src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
@@ -8,6 +8,8 @@ import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParseException;
 import com.google.gson.reflect.TypeToken;
+import com.suse.saltstack.netapi.datatypes.Arguments;
+import com.suse.saltstack.netapi.datatypes.Job;
 import com.suse.saltstack.netapi.datatypes.JobMinions;
 import com.suse.saltstack.netapi.datatypes.Keys;
 import com.suse.saltstack.netapi.datatypes.Token;
@@ -38,6 +40,8 @@ public class JsonParser<T> {
             new JsonParser<>(new TypeToken<Result<List<Token>>>(){});
     public static final JsonParser<Result<List<JobMinions>>> JOB_MINIONS =
             new JsonParser<>(new TypeToken<Result<List<JobMinions>>>(){});
+    public static final JsonParser<Result<List<Map<String, Job>>>> JOBS =
+            new JsonParser<>(new TypeToken<Result<List<Map<String, Job>>>>(){});
     public static final JsonParser<Result<List<Map<String, Object>>>> RETVALS =
             new JsonParser<>(new TypeToken<Result<List<Map<String, Object>>>>(){});
     public static final JsonParser<Stats> STATS =
@@ -58,6 +62,7 @@ public class JsonParser<T> {
         this.gson = new GsonBuilder()
                 .registerTypeAdapter(Date.class, new SaltStackDateDeserializer())
                 .registerTypeAdapter(Stats.class, new StatsDeserializer())
+                .registerTypeAdapter(Arguments.class, new ArgumentsDeserializer())
                 .create();
     }
 
@@ -119,6 +124,91 @@ public class JsonParser<T> {
                 return new Stats(app, server);
             } catch (Exception e) {
                 throw new JsonParseException(e);
+            }
+        }
+    }
+
+    /**
+     * Deserializer for Arguments class.
+     * Breaks the incoming arguments into args and kwargs parts
+     * and fills a new Arguments instance.
+     */
+    private class ArgumentsDeserializer implements JsonDeserializer<Arguments> {
+
+        private static final String KWARG_KEY = "__kwarg__";
+
+        @Override
+        public Arguments deserialize(JsonElement json, Type typeOfT,
+                JsonDeserializationContext context) throws JsonParseException {
+            Arguments result = new Arguments();
+
+            if (json != null && json.isJsonArray()) {
+                for (JsonElement jsonElement : json.getAsJsonArray()) {
+                    fillArgs(result, jsonElement);
+                }
+            }
+
+            return result;
+        }
+
+        /**
+         * Fills args/kwargs to given Arguments instance based on JSON data
+         * from jsonElement.
+         *
+         * @param result Arguments to be filled
+         * @param jsonElement input JSON data
+         */
+        private void fillArgs(Arguments result, JsonElement jsonElement) {
+            if (isKwarg(jsonElement)) {
+                filterKwarg(jsonElement);
+                fillKwargsFromObject(result, jsonElement.getAsJsonObject());
+            } else {
+                result.getArgs().add(gson.fromJson(jsonElement, Object.class));
+            }
+        }
+
+        /**
+         * Fills kwargs from given jsonObject key/values to given Arguments instance.
+         *
+         * @param result Arguments to be filled
+         * @param jsonObject input json data
+         */
+        private void fillKwargsFromObject(Arguments result, JsonObject jsonObject) {
+            for (Map.Entry<String, JsonElement> kwItem : jsonObject.entrySet()) {
+                result.getKwargs().put(kwItem.getKey(),
+                        gson.fromJson(kwItem.getValue(),
+                        Object.class));
+            }
+        }
+
+        /**
+         * Checks whether element is kwarg or arg.
+         * Element is a kwarg if it's an object and contains __kwarg__ property set to true.
+         *
+         * @param jsonElement element to be tested
+         * @return true if element is kwarg
+         */
+        private boolean isKwarg(JsonElement jsonElement) {
+            if (!jsonElement.isJsonObject()) {
+                return false;
+            }
+
+            JsonElement kwarg = jsonElement.getAsJsonObject().get(KWARG_KEY);
+            return kwarg != null
+                    && kwarg.isJsonPrimitive()
+                    && kwarg.getAsJsonPrimitive().isBoolean()
+                    && kwarg.getAsBoolean();
+        }
+
+        /**
+         * Filters __kwarg__ flag from given element.
+         *
+         * @param jsonElement
+         */
+        private void filterKwarg(JsonElement jsonElement) {
+            if (jsonElement.isJsonObject()
+                    && jsonElement.getAsJsonObject().has(KWARG_KEY)) {
+                jsonElement.getAsJsonObject().remove(KWARG_KEY);
             }
         }
     }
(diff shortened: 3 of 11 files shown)


src/main/java/com/suse/saltstack/netapi/datatypes/Job.java
```diff
@@ -1,17 +1,68 @@
...
+        return arguments;
     }
 
-    public List<String> getMinions() {
-        return minions;
+    public void setArguments(List<String> arguments) {
+        this.arguments = arguments;
     }
-}
+}
```
lucidd: Could you please add a newline here.

src/test/resources/jobs_response.json
```diff
@@ -0,0 +1,24 @@
...
+        ],
+        "Function": "test.echo",
+        "StartTime": "2015, Mar 04 20:01:10.485012",
+        "Target": "*",
+        "Target-type": "glob",
+        "User": "chuck"
+      }
+    }
+  ]
+}
```
lucidd: Here too :smile:

src/main/java/com/suse/saltstack/netapi/datatypes/Job.java
```diff
@@ -1,17 +1,68 @@
 package com.suse.saltstack.netapi.datatypes;
 
+import com.google.gson.annotations.SerializedName;
 import java.util.List;
 
+// todo - handle dates too (they are represented in really
```
lucidd: Minor detail but could you use the `TODO:` syntax for todos. Its more common and gets recognized by most IDEs and tools.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -314,6 +315,33 @@ public Job call() throws SaltStackException {
...
+                .getResult();
+         return result.getResult().get(0);
+    }
+
+    /**
+     * Get previously run jobs.
+     * @return future with a map containing run jobs keyed by job id.
+     * @throws SaltStackException if anything goes wrong
+     */
+    public Future<Map<String, Job>> getJobsAsync() throws SaltStackException {
```
lucidd: the `Exception` is not needed here. (also adjust docs)

src/main/java/com/suse/saltstack/netapi/datatypes/Job.java
```diff
@@ -1,17 +1,68 @@
...
+    private String user;
+
+    @SerializedName("Arguments")
+    private List<String> arguments;
+
+    public String getFunction() {
+        return function;
+    }
+
+    public void setFunction(String function) {
```
lucidd: Do you think we need `setters`? So far i don't think we need them and i like the idea of having the datatypes that come from the server immutable (even tho they are not completely right now).

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -5,9 +5,10 @@
 import com.suse.saltstack.netapi.config.ClientConfig;
 import static com.suse.saltstack.netapi.config.ClientConfig.*;
 import com.suse.saltstack.netapi.config.ProxySettings;
+import com.suse.saltstack.netapi.datatypes.Job;
```
lucidd: Minor nitpick but style changes / import reordering should be done in a separate commit/PR.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -314,6 +315,33 @@ public Job call() throws SaltStackException {
...
     /**
+     * Get previously run jobs.
+     * @return map containing run jobs keyed by job id.
+     * @throws SaltStackException if anything goes wrong
+     */
+    public Map<String, Job> getJobs() throws SaltStackException {
+         Result<List<Map<String, Job>>> result = connectionFactory
+                .create("/jobs", JsonParser.JOBS, config)
+                .getResult();
+         return result.getResult().get(0);
```
isbm: "Index out of range" crash here, once result is empty. This needs to be more safe. If this is all across everywhere that you need a first result, why not implement something like `result.getFirstResult()` instead?

src/test/java/com/suse/saltstack/netapi/parser/JsonParserTest.java
```diff
@@ -47,4 +50,37 @@ public void testSaltStackTokenParserWrongDate() throws Exception {
...
+
+        Map<String, Job> jobs = result.getResult().get(0);
+        Job job = jobs.get("20150304192951636258");
+        assertNotNull(job);
+        assertEquals(Arrays.asList(), job.getArguments());
+        assertEquals("test.ping", job.getFunction());
+        assertEquals("*", job.getTarget());
+        assertEquals("glob", job.getTargetType());
+        assertEquals("chuck", job.getUser());
+
```
isbm: Looks like that new line escaped from the end of the `Job.java` and wants to be moved there instead. :-)

lucidd:
After a little bit of testing it seems that the `Arguments` field is not just a `List` of `Strings` but a list of arbitrary `JsonElements`. It seems that if there are named parameters then they will always be in an object at the end of the list. The `__kwarg__`
seems to be always present in the named parameter object and was always `true` in my tests. 

Here is one of the results i got as an example:

```json
{
  "return": [
    {
      "20150306023815935637": {
        "Function": "pkg.install",
        "Target": "*",
        "Target-type": "glob",
        "Arguments": [
          "i3",
          true,
          {
            "sysupgrade": true,
            "__kwarg__": true
          }
        ],
        "StartTime": "2015, Mar 06 02:38:15.935637",
        "User": "lucid"
      }
    }
  ]
}
```

Generated by `salt '*' pkg.install i3 True sysupgrade=True`.

I think we will need a custom deserializer similar to the one in #44 to parse that.

src/main/java/com/suse/saltstack/netapi/datatypes/Job.java
renner: I agree with that @lucidd, we don't need the setters IIRC so let's have them removed.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
renner: Not too sure if it **needs** to be more safe, it actually depends on what the server is returning. If a helper method should be added then please via a separate PR.

isbm: If it is firmly guaranteed that there will be always one element, no matter what, then such code is OK. Otherwise nobody will like debug "Index out of range", while real error is "Server did not returned a proper data".

moio: @isbm it's not about being firmly guaranteed, it is about what is expected. We must make assumptions on what server returns based on its documented behavior - if it does not behave as it should, then it's OK for our code to behave in an undefined way, including failing with whatever `Exception` happens. We can't possibly foresee and wrap them all so it's better not to even attempt that and let the code fail "naturally" (examples: out of memory, hardware failures... will also generate cryptic stack traces, but they will be correct nonetheless).  Moreover `IndexOutOfRangeException` coming from here is pretty self-explanatory in this case.

If it is expected that sometimes there's no element, that's a whole different story of course.

isbm: Alright, @moio  there seems to be always at least one element, if I find this correctly:
https://github.com/saltstack/salt-api/blob/develop/saltapi/netapi/rest_cherrypy/app.py#L909

But I still find this pattern a bit ugly and `result.getFirst()` is a little bit better, IMO. :-)

hustodemon: Yes, I agree - getting the 1st element is ugly. I only followed this "pattern" because I saw it in different methods.

BTW - do you know if there is some declarative description of salt api (e.g. xsd file)? Or there is only python code in salt and the docs (which only describe couple of examples)? Having full api description in one place would be really terrific.

src/test/java/com/suse/saltstack/netapi/parser/JsonParserTest.java
hustodemon: Oh yeah, I've definitely problems with empty lines. (Maybe introducing checkstyle would be nice. It would save a lot of everyone's time and some reviewers' nerves).

hustodemon:
@lucidd thanks for the catch with args! I'll take a look at it.

src/main/java/com/suse/saltstack/netapi/datatypes/Job.java
hustodemon: yeees, but I'm too lazy to use shift :D out of curiosity:which ide do you use? (intellij idea works fine with it)

lucidd: Haha i know what you mean :) You could use intellijs live templates for that. I'm testing intellij myself right now but i know that eclipse does not recognize it as a todo. 
A relevant discussion [here](https://github.com/SUSE/saltstack-netapi-client-java/pull/44#discussion_r25929839) about todos. The TL;DR its preferred to open an issue to track todos and if we put todos in the code they should reference the issue as shown by @isbm in the discussion.

renner:
What is the current status of this, would you mind to rebase the patch to the current master? Do you further plan to fix the issues addressed in the inline comments?

hustodemon:
unfortunately I didn't have much time lately, but I'll definitely plan to finish this soon (I'll get to it this weekend).

src/main/java/com/suse/saltstack/netapi/datatypes/cherrypy/Arguments.java
```diff
@@ -0,0 +1,36 @@
+package com.suse.saltstack.netapi.datatypes.cherrypy;
```
lucidd: I think Arguments can go into datatypes directly the cherrypy is just for the stats stuff since that seems to be part of cherrypy and the netapi just taps into that.

src/test/resources/jobs_response.json
```diff
@@ -0,0 +1,24 @@
...
+        "Arguments": [],
+        "Function": "test.ping",
+        "StartTime": "2015, Mar 04 19:29:51.636258",
+        "Target": "*",
+        "Target-type": "glob",
+        "User": "chuck"
+      },
+      "20150304200110485012": {
+        "Arguments": [
+          "enable-autodestruction"
```
lucidd: I hope your system is still operational. :smile:

hustodemon: Autodestruction initiated by Chuck is lethal in most cases, but I was lucky and escaped it :D

src/test/resources/job_response.json
```diff
@@ -0,0 +1,27 @@
...
+                }
+            }
+        }
+    ],
+    "return": [
+        {
+            "blackbox": true
+        }
+    ]
+}
```
lucidd: missing newline here i'm not sure if checkstyle catches this yet for ressources (maybe something to add in #53)

hustodemon: yes, just added it.

2015-03-15 16:19 GMT+01:00 Kevin Walter <notifications@github.com>:

lucidd:
Sorry to bother you again but i think the Arguments parsing needs some more work i just did some testing with passing objects as parameters and currently there properties get put into `kwargs` even if they are passed as positional. Here are my findings so far:

* Positional arguments that are objects and contain a property "__kwarg__" with the value true get all merged together into the kwarg.

So to parse `kwarg` we need to look if the parameter is an object and if it contains a `__kwarg__` property with the value `true`. Otherwise it's part of  `arg`.


Here is an example that demonstrates that:

`salt '*' pkg.install '{refresh: true, __kwarg__: true}' '{sysupgrade: true, __kwarg__: true}' i3`

The `json` returned from `/jobs`.

```json
"20150315163041425361": {
        "Function": "pkg.install",
        "Target": "*",
        "Target-type": "glob",
        "Arguments": [
          {
            "refresh": true,
            "__kwarg__": true
          },
          {
            "sysupgrade": true,
            "__kwarg__": true
          },
          "i3"
        ],
        "StartTime": "2015, Mar 15 16:30:41.425361",
        "User": "lucid"
}
```

And here is the `pacman` log of what got executed so it did both refresh and sysupgrade which are `false` by default.

```
[2015-03-15 16:30] [PACMAN] Running 'pacman -S --noprogressbar --noconfirm --needed -y -u i3'          
[2015-03-15 16:30] [PACMAN] synchronizing package lists
[2015-03-15 16:30] [PACMAN] starting full system upgrade
```

There is also the `test.arg` function to experiment with arguments. It shows that for both
 `salt '*' test.arg i3 sysupgrade=true refresh=true` and `salt '*' test.arg '{refresh: true, __kwarg__: true}' '{sysupgrade: true, __kwarg__: true}' i3`
`kwargs` contains both sysupgrade and refresh with the value `true`.

hustodemon:
Thanks @lucidd for this. I updated the deserializer and added a new test.

hustodemon:
Oops, I think I pushed wrong version. Lemme re-check. Sorry for the noise!

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -119,4 +127,64 @@ public Stats deserialize(JsonElement jsonElement, Type type,
...
+         * as this is not needed further.
+         * @param jsonElement to be tested
+         * @return true if this element is a keyword argument
+         */
+        private boolean handleKwarg(JsonElement jsonElement) {
+            boolean kwarg = false;
+            if (jsonElement.isJsonObject()) {
+                JsonObject asJsonObject = jsonElement.getAsJsonObject();
+                kwarg = asJsonObject.has(KWARG_KEY)
+                        && Boolean.TRUE.equals(asJsonObject.get(KWARG_KEY).getAsBoolean());
```
lucidd: I think we should also test if  `__kwarg__` is a boolean before casting it since you can pass other types in `__kwarg__`.

Some personal preference stuff:

* I would drop the `Boolean.TRUE.equals`.
* Instead of using the `kwarg` variable just use if/else and return the proper values for each branch.
* I would prefer renaming this method to `isKwarg` and drop the removing of the `KWARG_KEY`. Instead just ignore the key in `fillKwargsFromObject`.

lucidd:
I think if you:

* Add some tests for the `getJobs` and `getJobsAsnyc`
* Address the boolean testing mentioned in the other comment.
* Rebase onto the latest master.
* Remove the WIP commit at the end. (should be part of another PR)

It should be ready for merging. :smile:

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
hustodemon: Hi,
* as for the `TRUE.equals` - it's there for null check reasons. This is handy when an evil user passes `__kwarg__: null`. We could also try to `getAsBoolean` and catch exceptions, but I think that's equally ugly :)
* handleKwarg stuff: if I put removing `__kwarg__` attribute to `fillKwargsFromObject`, it won't remove this attribute in case the passed arg is an object and has `__kwarg__` set to something else than `true`.
The question is: should we store `__kwarg__` in our representation of args/kwargs? I don't think so because this information is already contained - by placing attributes to `args`/`kwargs` attribute of `Arguments` class.

hustodemon:
I re-checked the handleKwargs method and broke it into two methods. This is more explicit and readable, I think...  feel free to take a look. Test pass ok, but I still need to do some manual testing tomorrow.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -122,4 +130,97 @@ public Stats deserialize(JsonElement jsonElement, Type type,
...
+         * @return true if element is kwarg
+         */
+        private boolean isKwarg(JsonElement jsonElement) {
+            if (!jsonElement.isJsonObject()) {
+                return false;
+            }
+
+            JsonObject asJsonObject = jsonElement.getAsJsonObject();
+
+            try {
```
lucidd: I agree with your comment from before that the boolean test with exception is ugly but there is a way around that `JsonPrimitive` has a `isBoolean` method. I had something like this in mind:
```java
private boolean isKwarg(JsonElement jsonElement) {
        if (jsonElement.isJsonObject()) {
            JsonObject jsonObject = jsonElement.getAsJsonObject();
            JsonElement kwarg =  jsonObject.get(KWARG_KEY);
        return kwarg != null
               && kwarg.isJsonPrimitive()
               && kwarg.getAsJsonPrimitive().isBoolean()
               && kwarg.getAsBoolean();
       } else {
            return false;
       }
}
```

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -122,4 +130,97 @@ public Stats deserialize(JsonElement jsonElement, Type type,
...
+         * from jsonElement.
+         *
+         * @param result Arguments to be filled
+         * @param jsonElement input JSON data
+         */
+        private void fillArgs(Arguments result, JsonElement jsonElement) {
+            boolean isKwarg = isKwarg(jsonElement);
+            // we don't need __kwarg__ attribute:
+            // args/kwargs are stored to separate attributes of Arguments class
+            // so we don't need this information anymore
```
lucidd: This is only partially true. Removing `__kwarg__` from objects that are actual kwarg objects is ok since the information is preserved by putting them in a different place in the `Arguments` class. But for normal objects that contain a `__kwarg__` with a value other than `true` (including other types) we lose the information of them being there and what value they had. That said judging from the results of the `test.arg` function the `__kwarg__` with values other than `true` gets dropped from the object and not passed to the function so i guess its ok not to preserve this information.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
hustodemon: YES! `getAsJsonPrimitive().isBoolean()` is the thing I was looking for, thanks :)

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
hustodemon: Yes, looking to `minion.py` it seems `__kwarg__` is discarded IIUC...

moio:
This PR has been open for quite some time and we see long-standing comments that haven't been addressed yet. Sorry @hustodemon, but we are closing it at the moment.

We hope this is only momentary though! If you have time please re-open it and fix all highlighted issues so that we can do a final review round and hopefully merge it into `master`.

Hope to hear from you soon!

hustodemon:
Hi @moio 
I believe all comments are addressed in my last patch, did I overlook something? 
Thanks.

moio:
I was probably misled by:
 - https://github.com/SUSE/saltstack-netapi-client-java/pull/45#discussion_r25895255: strange that GitHub still shows this one. I manually checked the file and looks OK;
 - https://github.com/SUSE/saltstack-netapi-client-java/pull/45#issuecomment-79068618: I saw no reply from you, is this OK now?
 - https://github.com/SUSE/saltstack-netapi-client-java/pull/45#issuecomment-84695103: same here.

Probably the problem is that we don't receive notifications when you `push -f`, so I was still looking at the old branch version with the `[WIP]` commit. Please make sure to say in comments when you are done with something so that we are up-to-date!

Will review this fully later today. Sorry for the noise.

hustodemon:
I see. I dealt with those comments (there was some discussion with @lucidd about the `__kwarg__` business but I think we reached consensus).

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -203,8 +204,8 @@ public Token call() throws SaltStackException {
      * @return object representing the scheduled job
      * @throws SaltStackException if anything goes wrong
      */
```
moio: Comment is out-of date: this method does not return the job id any more. Same for the asynchronous version.

hustodemon: It was wrong all the time. It seems that the comment itself is 1:1 copied from Salt API doc, which is wrong.

hustodemon: (anyway, I'll update this PR)

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
moio: @hustodemon do you mean [the pop call here](https://github.com/saltstack/salt/blob/develop/salt/minion.py#L264)?

In that case it is discarded, yes, but as an input to Salt. We are talking about a deserializer here, so we are parsing something from Salt.

How about moving line 166 to line 169, so that we are definitely sure we are not discarding anything useful? Would that work?

lucidd:
@moio I don't know why github still shows some of my comments. Last time i checked all my concerns got addressed so all comments of mine that are still visible should be invalid by now.

moio:
Apart from the two very minor points noted in line commits this is good to merge from my POV.

I would change the name `JobMinions` to something else, because in English it reads like "those minions who do a job" which is not really the intention here. But don't worry - since there's been so much activity on this PR already I prefer to merge it as it is and then I will fix the naming as a separate commit.

Thanks very much from your efforts, and sorry for this taking so long from our part. Hopefully communication will flow smoother next time!

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
hustodemon: done

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
hustodemon: Hi,
your suggestions makes sense to me, but I need to explore and test this a little bit further. Unfortunately I've been very busy lately so I'll take a closer look at the weekend. I'm sorry for this slowdown :(

moio: Don't worry, no need to rush - take your time to review my suggestions thoroughly :-)

hustodemon: OK, I played around with salt and I think you guys are right. It makes sense to preserve '__kwarg__' if the value is different from `true` as we could loose some information. It's changed in the last commit.

hustodemon:
I addressed your comments but I found another possible issue: multiple kwargs.

When you pass multiple kwarg to salt
`salt '*' test.arg mykwarg=1 mykwarg=2`
and then you list the job via the API, you get this:
```javascript
"20150404164512874404": {
    "Arguments": [
        {
            "__kwarg__": true,
            "mykwarg": 2
        }
    ],
    "Function": "test.arg",
    ...
}
```
You can see one of the kwargs is swallowed.

But when you are evil and pass the kwarg as arg:
`sudo salt '*' test.arg mykwarg=1 '{__kwarg__: true, mykwarg: 123}`
, the `Arguments` section looks like this:
```javascript
"Arguments": [
    {
        "__kwarg__": true,
        "mykwarg": 123
    },
    {
        "__kwarg__": true,
        "mykwarg": 1
    }
]
```
which is confusing. I thought such thing like multiple kwarg doesn't exist :) The question is if this is bug or a feature, maybe asking SaltStack people would be way to go.

My patch doesn't support parsing multiple kwarg which is wrong IMHO (I added a simple test that documents that anyway). If we want to support it, we have 2 options:
1. Change `Arguments.kwargs` to multimap (e.g. `Map<String, List<Object>>`)
2. handle every block of Arguments JSON that has `__kwarg__: true` independently, that would mean `Arguments.kwargs` would be `List<Map<String, Object>>`.

I'd personally go with 2, what do you think?

Thanks for the review.

lucidd:
I think the whole manipulating `__kwarg__` manually is hacky anyway and not how it would be used in practice. As far as i understand there a few things here to differentiate.

* The way the `salt` command works i.e what it sends over the wire.
* What gets received by the master. (this is what gets logged and returned by `/jobs`)
* What gets passed to the function that got called. (this is what gets returned calling the `test.arg` function.

What you looked at was the result of `/jobs` which is the raw arguments as they got received by the master. 
I think what happens here is that the `salt` command just sends the last value it gets for a certain argument name so the first one does not show up in the raw arguments.

`salt '*' test.arg mykwarg=1 mykwarg=2 --out json`

```json
{
    "lullabye": {
        "args": [], 
        "kwargs": {
            "__pub_user": "lucid", 
            "__pub_arg": [
                {
                    "mykwarg": 2
                }
            ], 
            "__pub_fun": "test.arg", 
            "__pub_jid": "20150404213432177251", 
            "__pub_tgt": "*", 
            "__pub_tgt_type": "glob", 
            "__pub_ret": "", 
            "mykwarg": 2
        }
    }
}
```

If you look at the `test.arg` output of your second example it behaves the same and only has the `mykwarg` with the value `1` in `kwargs`. There is still the other one in `__pub_arg` but i think that's not what a normal function would look at.

`salt '*' test.arg mykwarg=1 '{__kwarg__: true, mykwarg: 123}' --out json`

```json
{
    "lullabye": {
        "args": [], 
        "kwargs": {
            "__pub_user": "lucid", 
            "__pub_arg": [
                {
                    "mykwarg": 123
                }, 
                {
                    "mykwarg": 1
                }
            ], 
            "__pub_fun": "test.arg", 
            "__pub_jid": "20150404212534055290", 
            "__pub_tgt": "*", 
            "__pub_tgt_type": "glob", 
            "__pub_ret": "", 
            "mykwarg": 1
        }
    }
}
```
What our `Arguments` class models right know is the parsed version of the arguments as received by the function. This would be what most people would want to access.

I think if we go that route and preserve all information that's present in the raw arguments from `/jobs` then the best way to do it is to just keep them unparsed in a separate field.

The only use case i see for them is for debugging and logging since it could help with troubleshooting problems.

hustodemon:
Your suggestions make sense to me and it would be easiest and _correct_ way to store all the args to string.

> The only use case i see for them is for debugging and logging since it could help with troubleshooting problems.

Originally I meant args/kwargs to be also used for use case like listing job and rerunning it with same arguments/slightly changed arguments. With all args/kwargs as strings it is still available but not so convenient I'd say. On the other hand it seems we don't have a lot of options with salt shuffling args/kwargs :)

I'll think about it a bit more and if nothing crosses my mind, I'll go with your "string args".

Thanks a lot!

lucidd:
Sorry it seems like i made it a little bit hard to understand what i mean. What i had in mind was something like this.

```java
class Arguments {
  private JsonArray raw;
  private List<Object> args;
  private Map<String, Object> kwargs;
}
```

This way we still have the `args` and `kwargs` as most people would use it but also give access to the `raw` arguments which preserves information about how the arguments were send (multiple `__kwarg__` and all of that madness).
This information is what i meant was most likely only useful for debugging and logging.

I hope that clears i up a bit :smile: 
If there is anything unclear or i got something wrong just let me know.

hustodemon:
Ah, I thought you meant replacing `args`/`kwargs` :)

Just two "philosofic" points:
1. I'm just tentative about duplicating the information in that class. Having the information there twice can have weird consequences - the class is immutable and if a user of this it changes something in `raw` field, it won't be reflected to `args`/`kwargs`. Maybe we could only have raw data and provide some kind of utility to convert it to `args`/`kwargs`. 

2. (Also, until now, we didn't have "GSON" field in our `datatypes`. It's nice that it provides neat methods for analyzing and manipulating underlying data but on the other hand it'd leak GSON outside of the client. Not sure if that's a good thing, maybe we should expose just `Object` and let the user handle this.).

Anyway, I'll have computer access in the evening again so I'll try to play around with it to see what it looks like. Then I'll push a new version.

Thanks!

lucidd:
Good points.

* I think we should clearly document that Arguments is supposed to be immutable and we don't guarantee the behavior if someone changes something. (i think that's the case for most of our datatypes.)

* I don't think its a good thing to leak the gson datatypes but we don't have much choice other than writing our own abstraction and that wouldn't be any better than what gson already provides.

I think it would be best to merge this PR without the `raw` arguments and discuss that in an issue since the `raw` arguments have very limited usage and are not that important for regular use.

moio:
Guys, thanks for the detailed discussion.

Regardless of the final results, together with @isbm we agreed that current code is definitely good enough for merging and that we don't necessarily want to wait until all those tiny details are perfect. Current code is definitely good enough for inclusion.

Thus I am merging this PR in its current state, right now. As always feel free to improve the code base with other PRs or start discussions in issues.

Thank you again!!!

hustodemon:
Ok, I'll focus on the follow-up patch then (adding result to a single job listing - it was previously marked as WIP).

Many thanks to everyone!!

PR #46, opened 2015-03-06
Add a config parameter CONNECTION_TIMEOUT

xavierfacq:
Add a config parameter CONNECTION_TIMEOUT to set the maximum connection timeout when performing HTTP requests.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/client/impl/JDKConnection.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/impl/JDKConnection.java
@@ -75,10 +75,14 @@ public class JDKConnection<T> implements Connection<T> {
 
             // Timeouts may be specified on configuration
             int connect_timeout = config.get(CONNECT_TIMEOUT);
-            if (connect_timeout >= 0) connection.setConnectTimeout(connect_timeout);
+            if (connect_timeout >= 0) {
+                connection.setConnectTimeout(connect_timeout);
+            }
 
             int socket_timeout = config.get(SOCKET_TIMEOUT);
-            if (socket_timeout >= 0) connection.setReadTimeout(socket_timeout);
+            if (socket_timeout >= 0) {
+                connection.setReadTimeout(socket_timeout);
+            }
 
             // Send data in case we have some
             if (data != null) {
--- a/src/main/java/com/suse/saltstack/netapi/client/impl/JDKConnection.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/impl/JDKConnection.java
@@ -2,6 +2,7 @@ package com.suse.saltstack.netapi.client.impl;
 
 import com.suse.saltstack.netapi.client.Connection;
 import com.suse.saltstack.netapi.config.ClientConfig;
+import static com.suse.saltstack.netapi.config.ClientConfig.*;
 import com.suse.saltstack.netapi.exception.SaltStackException;
 import com.suse.saltstack.netapi.parser.JsonParser;
 
@@ -69,6 +70,12 @@ public class JDKConnection<T> implements Connection<T> {
             connection.setUseCaches(false);
             connection.setDoInput(true);
             connection.setDoOutput(true);
+
+            int connectionTimeout = config.get(JDKC_CONNECTION_TIMEOUT);
+            if (connectionTimeout >= 0) connection.setConnectTimeout(connectionTimeout);
+
+            int readTimeout = config.get(JDKC_READ_TIMEOUT);
+            if (readTimeout >= 0) connection.setReadTimeout(readTimeout);
 
             // Send data in case we have some
             if (data != null) {
--- a/src/main/java/com/suse/saltstack/netapi/client/impl/JDKConnection.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/impl/JDKConnection.java
@@ -2,7 +2,9 @@ package com.suse.saltstack.netapi.client.impl;
 
 import com.suse.saltstack.netapi.client.Connection;
 import com.suse.saltstack.netapi.config.ClientConfig;
+
 import static com.suse.saltstack.netapi.config.ClientConfig.*;
+
 import com.suse.saltstack.netapi.exception.SaltStackException;
 import com.suse.saltstack.netapi.parser.JsonParser;
 
@@ -71,11 +73,9 @@ public class JDKConnection<T> implements Connection<T> {
             connection.setDoInput(true);
             connection.setDoOutput(true);
 
-            int connectionTimeout = config.get(REQUEST_CONNECTION_TIMEOUT);
-            if (connectionTimeout >= 0) connection.setConnectTimeout(connectionTimeout);
-
-            int readTimeout = config.get(REQUEST_CONNECTION_TIMEOUT);
-            if (readTimeout >= 0) connection.setReadTimeout(readTimeout);
+            // Timeouts may be specified on configuration
+            connection.setConnectTimeout(config.get(CONNECT_TIMEOUT));
+            connection.setReadTimeout(config.get(SOCKET_TIMEOUT));
 
             // Send data in case we have some
             if (data != null) {
--- a/src/main/java/com/suse/saltstack/netapi/client/impl/JDKConnection.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/impl/JDKConnection.java
@@ -74,8 +74,11 @@ public class JDKConnection<T> implements Connection<T> {
             connection.setDoOutput(true);
 
             // Timeouts may be specified on configuration
-            connection.setConnectTimeout(config.get(CONNECT_TIMEOUT));
-            connection.setReadTimeout(config.get(SOCKET_TIMEOUT));
+            int connect_timeout = config.get(CONNECT_TIMEOUT);
+            if (connect_timeout >= 0) connection.setConnectTimeout(connect_timeout);
+
+            int socket_timeout = config.get(SOCKET_TIMEOUT);
+            if (socket_timeout >= 0) connection.setReadTimeout(socket_timeout);
 
             // Send data in case we have some
             if (data != null) {
--- a/src/main/java/com/suse/saltstack/netapi/client/impl/JDKConnection.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/impl/JDKConnection.java
@@ -71,10 +71,10 @@ public class JDKConnection<T> implements Connection<T> {
             connection.setDoInput(true);
             connection.setDoOutput(true);
 
-            int connectionTimeout = config.get(JDKC_CONNECTION_TIMEOUT);
+            int connectionTimeout = config.get(REQUEST_CONNECTION_TIMEOUT);
             if (connectionTimeout >= 0) connection.setConnectTimeout(connectionTimeout);
 
-            int readTimeout = config.get(JDKC_READ_TIMEOUT);
+            int readTimeout = config.get(REQUEST_CONNECTION_TIMEOUT);
             if (readTimeout >= 0) connection.setReadTimeout(readTimeout);
 
             // Send data in case we have some
--- a/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
+++ b/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
@@ -12,8 +12,8 @@ public class ClientConfig {
     public static final Key<URI> URL = new Key<>(URI.create("http://localhost:8000"));
     public static final Key<String> TOKEN = new Key<>();
 
-    // Timeout settings setted at 1000ms (10s)
-    public static final Key<Integer> CONNECTION_TIMEOUT = new Key<>(1000);
+    // Timeout settings setted at 10000ms (10s)
+    public static final Key<Integer> CONNECTION_TIMEOUT = new Key<>(10000);
 
     // Proxy settings
     public static final Key<String> PROXY_HOSTNAME = new Key<>();
--- a/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
+++ b/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
@@ -20,7 +20,16 @@ public class ClientConfig {
      * @see HttpClientConnection.request(String)
      * @see JDKConnection.request(String, String)
      */
-    public static final Key<Integer> REQUEST_CONNECTION_TIMEOUT = new Key<>(10000);
+    public static final Key<Integer> CONNECT_TIMEOUT = new Key<>(10000);
+
+    /**
+     * Sets a specified timeout value, in milliseconds, to be used when waiting for data in HttpClientConnection / JDKConnection.
+     * A timeout of zero is interpreted as an infinite timeout. Default value is 10000ms (10s).
+     * 
+     * @see HttpClientConnection.request(String)
+     * @see JDKConnection.request(String, String)
+     */
+    public static final Key<Integer> SOCKET_TIMEOUT = new Key<>(10000);
 
     // Proxy settings
     public static final Key<String> PROXY_HOSTNAME = new Key<>();
--- a/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
+++ b/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
@@ -14,8 +14,8 @@ public class ClientConfig {
 
     // Request timeout settings
     /**
-     * Sets a specified timeout value, in milliseconds, to be used when dealing a communications link in HttpClientConnection / JDKConnection.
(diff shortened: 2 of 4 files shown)


src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
```diff
@@ -12,6 +12,9 @@
     public static final Key<URI> URL = new Key<>(URI.create("http://localhost:8000"));
     public static final Key<String> TOKEN = new Key<>();
 
+    // Timeout settings setted at 1000ms (10s)
+    public static final Key<Integer> CONNECTION_TIMEOUT = new Key<>(1000);
```
dgiagio: Good job on the patch. But isn't 1000ms equal to 1s instead of 10s?

xavierfacq: Thanks, there was a mistake in my units, it's commited.

dgiagio:
@xavierfacq Ok, good. Can you also add a unit test for the timeout? The WireMock framework has some options to delay the response, so you can ensure it's working the way it's supposed to.

lucidd:
@xavierfacq Thanks for your Contribution! 
Would you mind implementing the timeout also for the `JDKConnection` backend? [Here](http://docs.oracle.com/javase/7/docs/api/java/net/URLConnection.html#setConnectTimeout(int)) is the relevant documentation.

xavierfacq:
@dgiagio I'll add the testunit ASAP
@lucidd Is the JDKConnection implementation ok for you ?

lucidd:
@xavierfacq Im not sure yet i will have to look into it more. I think the best case would be to have the same config keys used by both back ends and have them behave the same to avoid confusion. I'm not sure if this is possible tho (also considering possibly having additional back ends in the future). What do you think would that be possible?

xavierfacq:
@lucidd You are right, I'll simplify and add only a new key named REQUEST_CONNECTION_TIMEOUT,   used by both back ends.

lucidd:
I looked into both connection back ends a bit and it seems that there are 2 types of timeouts that both back ends can provide. 

            | JDK          | HttpClient
-----------|------------- | -------------
initial connection | [setConnectTimeout](http://docs.oracle.com/javase/7/docs/api/java/net/URLConnection.html#setConnectTimeout(int))  | [setConnectTimeout](http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/client/config/RequestConfig.html#getConnectTimeout())
waiting for data | [setReadTimeout](http://docs.oracle.com/javase/7/docs/api/java/net/URLConnection.html#setReadTimeout(int))  | [setSocketTimeout](http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/client/config/RequestConfig.html#getSocketTimeout())

Note: i linked the getters for `HttpClient` since only they have documentation.

So i would suggest having 2 config keys `CONNECT_TIMEOUT` and `SOCKET_TIMEOUT` that are used by both back ends.

xavierfacq:
Ok, I'll omit the setConnectionRequestTimeout() and only set the two main timeouts : 
- opening a communications link : using setConnectTimeout()
- waiting for data : using setReadTimeout() / setSocketTimeout()

Ok for your config key names ;-)

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
```diff
@@ -146,6 +148,23 @@ public void testRunRequestAsync() throws Exception {
                 .withRequestBody(equalToJson(JSON_RUN_REQUEST)));
     }
 
+    @Test(expected = SaltStackException.class)
+    public void testRunRequestWithTimeout() throws Exception {
+    	// create a local SaltStackClient with a fast timeout configuration to do not lock tests more thant 2s
```
lucidd: here is a tab that tried to sneak in ;)

src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
```diff
@@ -12,6 +12,25 @@
     public static final Key<URI> URL = new Key<>(URI.create("http://localhost:8000"));
     public static final Key<String> TOKEN = new Key<>();
 
+    // Request timeout settings
+    /**
+     * Sets a specified timeout value, in milliseconds, to be used when dealing a communications link in HttpClientConnection / JDKConnection.
+     * A timeout of zero is interpreted as an infinite timeout. Default value is 10000ms (10s).
```
lucidd: I would write something more similar to the HttpClient documentation:
`Timeout in milliseconds until a connection is established. A timeout of zero is interpreted as an infinite timeout. Default value is 10000ms (10s)`

src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
```diff
@@ -12,6 +12,25 @@
...
+     * 
+     * @see HttpClientConnection.request(String)
+     * @see JDKConnection.request(String, String)
+     */
+    public static final Key<Integer> CONNECT_TIMEOUT = new Key<>(10000);
+
+    /**
+     * Sets a specified timeout value, in milliseconds, to be used when waiting for data in HttpClientConnection / JDKConnection.
+     * A timeout of zero is interpreted as an infinite timeout. Default value is 10000ms (10s).
```
lucidd: And here too:
`Timeout in milliseconds for waiting for data. A timeout of zero is interpreted as an infinite timeout. Default value is 10000ms (10s)`

lucidd:
@xavierfacq We should test both timeouts separately and with both back ends. You can pass the back end to use to the `SaltStackClient`.
That said i haven't found a way to test `CONNECT_TIMEOUT` yet since `withFixedDelay`
does not work in that case. Maybe @dgiagio knows more.

xavierfacq:
I have changed the Javadoc and added the following text : `A negative value is interpreted as undefined (system default).` I have modified the code of JDKConnection to deal with negative values, to be compliant with HttpClient.

I'm looking for a solution to test the CONNECT_TIMEOUT in SaltStackClientTest, and I'll create tests for JDKConnection.

src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
```diff
@@ -12,6 +12,25 @@
     public static final Key<URI> URL = new Key<>(URI.create("http://localhost:8000"));
     public static final Key<String> TOKEN = new Key<>();
 
+    // Request timeout settings
```
renner: I don't think we need this single line comment since there is the javadoc one below.

src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
```diff
@@ -12,6 +12,25 @@
...
+     * Timeout in milliseconds until a connection is established. A timeout of zero is interpreted as an infinite timeout. A negative value is interpreted as undefined (system default).
+     * Default value is 10000ms (10s)
+     * 
+     * @see HttpClientConnection.request(String)
+     * @see JDKConnection.request(String, String)
+     */
+    public static final Key<Integer> CONNECT_TIMEOUT = new Key<>(10000);
+
+    /**
+     * Timeout in milliseconds for waiting for data. A timeout of zero is interpreted as an infinite timeout. A negative value is interpreted as undefined (system default).
```
renner: Please follow the maximum line length convention (92 characters) also in javadoc comments, here and above.

src/main/java/com/suse/saltstack/netapi/client/impl/JDKConnection.java
```diff
@@ -70,6 +73,13 @@ private T request(String method, String data)
             connection.setDoInput(true);
             connection.setDoOutput(true);
 
+            // Timeouts may be specified on configuration
+            int connect_timeout = config.get(CONNECT_TIMEOUT);
+            if (connect_timeout >= 0) connection.setConnectTimeout(connect_timeout);
```
isbm: No, please don't.
http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-142311.html

src/main/java/com/suse/saltstack/netapi/client/impl/JDKConnection.java
```diff
@@ -70,6 +73,13 @@ private T request(String method, String data)
...
+
+            int socket_timeout = config.get(SOCKET_TIMEOUT);
+            if (socket_timeout >= 0) connection.setReadTimeout(socket_timeout);
```
isbm: Add braces here as well.

src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
```diff
@@ -12,6 +12,28 @@
...
+    /**
+     * Timeout in milliseconds for waiting for data.
+     * A timeout of zero is interpreted as an infinite timeout.
+     * A negative value is interpreted as undefined (system default).
+     * Default value is 10000ms (10s)
+     * 
+     * @see HttpClientConnection.request(String)
+     * @see JDKConnection.request(String, String)
+     */
+    public static final Key<Integer> SOCKET_TIMEOUT = new Key<>(10000);
```
isbm: Please move `10000` to a `DEFAULT_TIMEOUT` variable.

src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
```diff
@@ -12,6 +12,28 @@
...
+    /**
+     * Timeout in milliseconds until a connection is established.
+     * A timeout of zero is interpreted as an infinite timeout.
+     * A negative value is interpreted as undefined (system default).
+     * Default value is 10000ms (10s)
+     * 
+     * @see HttpClientConnection.request(String)
+     * @see JDKConnection.request(String, String)
+     */
+    public static final Key<Integer> CONNECT_TIMEOUT = new Key<>(10000);
```
isbm: Please see comment for line 35.

src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
```diff
@@ -12,6 +12,28 @@
...
+     * Default value is 10000ms (10s)
+     * 
+     * @see HttpClientConnection.request(String)
+     * @see JDKConnection.request(String, String)
+     */
+    public static final Key<Integer> SOCKET_TIMEOUT = new Key<>(10000);
+
     // Proxy settings
     public static final Key<String> PROXY_HOSTNAME = new Key<>();
     public static final Key<Integer> PROXY_PORT = new Key<>(3128);
```
isbm: If this port is not a random idea, I would suggest to move it to constants instead.

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
```diff
@@ -147,6 +155,46 @@ public void testRunRequestAsync() throws Exception {
...
     @Test
+    public void testRunRequestWithSocketTimeout() throws Exception {
+        exception.expect(SaltStackException.class);
+        exception.expectMessage(containsString("Read timed out"));
+
+        // create a local SaltStackClient with a fast timeout configuration 
+        // to do not lock tests more thant 2s
+        URI uri = URI.create("http://localhost:" + Integer.toString(MOCK_HTTP_PORT));
+        SaltStackClient clientWithFastTimeout = new SaltStackClient(uri);
+        clientWithFastTimeout.getConfig().put(SOCKET_TIMEOUT, 1000);
```
isbm: Bug. You use default timeout 10 seconds, but here only one second.

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
```diff
@@ -147,6 +155,46 @@ public void testRunRequestAsync() throws Exception {
...
+    @Test
+    public void testRunRequestWithSocketTimeoutThroughJDKConnection() throws Exception {
+        exception.expect(SaltStackException.class);
+        exception.expectMessage(containsString("Read timed out"));
+
+        // create a local SaltStackClient with a fast timeout configuration 
+        // to do not lock tests more thant 2s
+        URI uri = URI.create("http://localhost:" + Integer.toString(MOCK_HTTP_PORT));
+        SaltStackClient clientWithFastTimeout = new SaltStackClient(uri, new JDKConnectionFactory());
+        clientWithFastTimeout.getConfig().put(SOCKET_TIMEOUT, 1000);
```
isbm: Same here, I guess. I would strongly suggest to use `DEFAULT_TIMEOUT` constant instead.

src/main/java/com/suse/saltstack/netapi/client/impl/JDKConnection.java
isbm: Also, I am not sure if we want to trim negative here. If `config` is expecting to return 0, then it setting timeout to zero is same as disabling it, so you can just set it directly from getting it. But if configuration has negative value, this should be as configuration error, while this information is lost at this point.

But I don't have a strong opinion here. @renner, @moio ?

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
renner: Why is this a bug @isbm? He suggests to use 1 second here (and below) to not block the tests for 10 seconds. Makes sense to me actually.

isbm: Because you do not test a default timeout? :-) In this case you can put 50 ms as well, as a positive non-zero will set a timeout as per se.

src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
renner: This port (3128) was chosen as the default PROXY_PORT since it is the default port used by some HTTP proxy servers, e.g. squid. Making this a constant? Why not, but no need to discuss this here since it is very unrelated to this patch.

isbm: @renner agreed.

src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
xavierfacq: Is it ok for you if I add a `private static final int DEFAULT_TIMEOUT = 10000;` used to initialize `CONNECT_TIMEOUT` and `SOCKET_TIMEOUT` ?

isbm: Why constant `DEFAULT_TIMEOUT` has to be private?

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
xavierfacq: This two tests check that the "Read timed out" is well implemented, that's why I force a short timeout of 1s, and delay the server response of 2s, to catch and valid the "Read timed out" error. This is just an hack to quickly pass tests. If we want to test with the default config, we have to delay the response more than 10000ms, twice... 

Is it OK ?

src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
xavierfacq: There are already two public config keys, a third one controlling the two others will be strange, no ?

isbm: It is already strange we have so many default timeouts, is it?

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
renner: I agree and from my perspective this test is OK as it is using the 1 second timeout. It could in theory even be smaller as @isbm suggested, but IMO 1 second is small enough. The default value could maybe be verified just like this (if this is even necessary)?

    assertEquals(10000, new SaltStackClient().getConfig().get(SOCKET_TIMEOUT));

src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
renner: I would say this is not necessary and can stay as it is and @moio agrees with that as well.

lucidd: I don't think we need extra constants for the default values of keys since they are already pretty self explanatory constants accessible through `SOCKET_TIMEOUT.defaultValue`.
Besides that they may have the same default value now but that could change.

renner:
I merged it, thanks for your work so far! Please have a look at the fixes that I added on top of your commits, especially those two:

- Give proper names to variables (7bc990ed5bcfe44e44c471303cef791ddc777640)
- Fix typos in comments (9be2854917fbeac0d46087ea6cca646a7e904981)

xavierfacq:
I agree for your fixes, thanks for the merge !

PR #48, opened 2015-03-06
Cleanup and style adjustments

lucidd:
This PR tries to unify the formatting a bit more.
I adopted the style that was most common across the project so far.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/client/impl/HttpClientConnection.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/impl/HttpClientConnection.java
@@ -71,7 +71,6 @@ public class HttpClientConnection<T> implements Connection<T> {
     /**
      * Perform HTTP request and parse the result into a given result type.
      *
-     * @param method the HTTP method to use
      * @return object of type given by resultType
      * @throws SaltStackException in case of a problem
      */
--- a/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
+++ b/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
@@ -85,6 +85,6 @@ public class ClientConfig {
     @SuppressWarnings("unchecked")
     public <T> T get(Key<T> key) {
        Object value = store.get(key);
-       return value != null ? (T)value : key.defaultValue;
+       return value != null ? (T) value : key.defaultValue;
     }
 }
--- a/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
+++ b/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
@@ -86,5 +86,4 @@ public class ClientConfig {
        Object value = store.get(key);
        return value != null ? (T)value : key.defaultValue;
     }
-
 }
--- a/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
+++ b/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
@@ -74,12 +74,13 @@ public class ClientConfig {
     }
 
     /**
-     * Returns the configured value for the given key. If the key is not explicitly set. the
-     * default value is for that key is returned.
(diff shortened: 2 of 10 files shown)


src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
```diff
@@ -86,5 +86,4 @@ public Key(T defaultValue) {
        Object value = store.get(key);
        return value != null ? (T)value : key.defaultValue;
```
renner: While we are at doing style adjustments: I usually prefer to put a blank in between the type and the `value` that should be casted. Feel free to add a commit to fix that as well :wink:

src/test/java/com/suse/saltstack/netapi/config/ClientConfigTest.java
```diff
@@ -31,5 +31,4 @@ public void testPutGetRemove() {
         config.remove(key);
         assertEquals("Should return the default value after removing the key", config.get(key), key.defaultValue);
```
isbm: The part `key.default....` probably better start from a new line.

src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
lucidd: That feels weird but i will include it. If there are any other styling issues you see let me know. :)

src/test/java/com/suse/saltstack/netapi/config/ClientConfigTest.java
lucidd: I think i will wrap after the description like so:
```java
assertEquals("Should return the default value after removing the key",
        config.get(key), key.defaultValue);
```
That way it fits the 92 limit and i think it looks nice having `expected` and `actual` value together. Is that ok with you?

src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
renner: It is mostly my personal preference it seems, I found examples doing it with no blank after the `)` as well, just like above. However, at least at [google] (https://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s4.6.2-horizontal-whitespace) they seem to put blanks there (see 5.).

src/test/java/com/suse/saltstack/netapi/config/ClientConfigTest.java
renner: Yes, that's good, thanks.

src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
lucidd: I usually use it like i would other unary operators but like you said its just personal preference. I don't have a problem with putting a whitespace there just need to get used to it. :)

lucidd:
Ok this patch now includes style adjustments to enforce the 92 character limit throughout the project (i hope i didn't miss a case). And the whitespace after a cast proposed by @renner. If there is any other styling issue that bugs you please let me know.

src/main/java/com/suse/saltstack/netapi/client/impl/HttpClientConnection.java
```diff
@@ -71,7 +71,6 @@ public T getResult() throws SaltStackException {
     /**
      * Perform HTTP request and parse the result into a given result type.
      *
-     * @param method the HTTP method to use
```
moio: I guess we should add `@param data` here.

renner: Correct, @moio, we didn't enable the `JavadocMethod` checkstyle module yet since it requires a bigger patch. Anyways this is fixed with f0ad98c736387151e5937efdab8eb4ebf368b61e.

renner:
Closing since all of this is now in master after we merged #50.

PR #50, opened 2015-03-10
Checkstyle

hustodemon:
The PR adds checkstyle to the project which should spare some time wasted on reviews (about complaining about code formatting, conventions etc.).

The first patch makes the code ready for checkstyle to pass.
The second one integrates checkstyle itself. I used a minimalistic checkstyle configuration (basically only rules that conform to coding guidelines in README plus a couple of useful modules (let me know if you want me to remove some of them)).

If you decide to integrate checkstyle, I think it'd be useful to adjust Travis CI to execute checkstyle goal as well.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
@@ -387,32 +387,32 @@ public class SaltStackClient {
         return executor.submit(callable);
     }
 
-  /**
-  * Query statistics from the CherryPy Server.
-  *
-  * GET /stats
-  *
-  * @return The {@link Stats} object.
-  * @throws SaltStackException if anything goes wrong
-  */
-  public Stats stats() throws SaltStackException {
-      return connectionFactory.create("/stats", JsonParser.STATS, config).getResult();
-  }
+    /**
+     * Query statistics from the CherryPy Server.
+     *
+     * GET /stats
+     *
+     * @return The {@link Stats} object.
+     * @throws SaltStackException if anything goes wrong
+     */
+    public Stats stats() throws SaltStackException {
+        return connectionFactory.create("/stats", JsonParser.STATS, config).getResult();
+    }
 
-  /**
-  * Asynchronously query statistics from the CherryPy Server.
-  *
-  * GET /stats
-  *
-  * @return Future containing the {@link Stats} object.
-  */
-  public Future<Stats> statsAsync() {
-      Callable<Stats> callable = new Callable<Stats>() {
-          @Override
-          public Stats call() throws SaltStackException {
-              return stats();
-          }
-      };
-      return executor.submit(callable);
-  }
+    /**
+     * Asynchronously query statistics from the CherryPy Server.
+     *
+     * GET /stats
+     *
+     * @return Future containing the {@link Stats} object.
+     */
+    public Future<Stats> statsAsync() {
+        Callable<Stats> callable = new Callable<Stats>() {
+            @Override
+            public Stats call() throws SaltStackException {
+                return stats();
+            }
+        };
+        return executor.submit(callable);
+    }
 }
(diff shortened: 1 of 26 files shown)


dgiagio:
It's a very good idea IMO to use checkstyle. Thanks for the patch! I hope it gets merged.

renner:
Yes, I like to have checkstyle and would merge such a patch. Could you add a commit to also modify `.travis.yml` file in order to run checkstyle before every build? This patch should conflict with the changes proposed in #48 and actually should make pull request #48 obsolete. At least I would expect all the changes that are done there also to happen here, if this is not the case we might be missing some rules in `checkstyle.xml`.

Here is some rules that should probably be added there as well:

- `EmptyLineSeparator`
- `JavadocType`
- `JavadocMethod` (at least for public methods)
- `WhitespaceAround`

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -387,32 +387,32 @@ public Job call() throws SaltStackException {
...
+     * @return The {@link Stats} object.
+     * @throws SaltStackException if anything goes wrong
+     */
+    public Stats stats() throws SaltStackException {
+        return connectionFactory.create("/stats", JsonParser.STATS, config).getResult();
+    }
+
+    /**
+     * Asynchronously query statistics from the CherryPy Server.
+     * <p/>
```
renner: Fixing indentation of this code is fine, but why there is empty paragraphs (`<p/>`) added here and above?

lucidd:
Awesome thanks for this! This will save so much time :).
I agree with @renner use #48 as reference and once everything is included here i will close it.

hustodemon:
Hi @lucidd , I think your patch PR #48 looks good should be merged first, shouldn't it?

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
hustodemon: Hmm, probably the IDE put it there, I don't remember typing it there manually. Will fix in next push.

renner:
@hustodemon: everything that @lucidd had changed for #48 should be pointed out by checkstyle, which is a very good test for your rules! It would therefore be good if you could first configure checkstyle so that it complains about all those lines, and then satisfy the rules by doing the changes in here as `Fix code-convention errors`. Once all of his changes are here we know that the rules are good and we will just close #48 and merge only this PR.

hustodemon:
No problem, I'll include it to my PR. I'll push it this evening.

hustodemon:
Ok, I pushed following things:
- @lucidd style-correcting patches
- adjustments suggested by @renner 
- checkstyle patch itself (dependency in pom.xml, checkstyle.xml config, .travil.yml adjustment to run checkstyle:check goal on push)
I didn't do following:
-  didn't add the JavadocMethod module - for public methods it causes like 65 checkstyle violations. I suggest doing that in a separate PR. Or I can add it to this one, it just won't be today.
- there are some TODOs is JavadocType patch. Some javadocs for classes are missing - I couldn't find any reasonable documentation to the api and salt rest_cherrypy reference is not really good. Could anyone point me to reasonable documentation (I just don't want to write any nonsense/inaccurate things to javadoc).
Looking forward to your feedback!

renner:
I merged it (skipping only 18e2dc1) and did some more adjustments on top:

- Allow empty types in WhitespaceAround (c47f941eec0b83f291c1f077a4eabbd54723176c)
- Ignore maximum line length for imports (d2f80f16750fcd7405e82630f8344e77f5f0ff07)
- Avoid star imports (b8620d51a1d3af963caf88c60216a65d25dda8df)
- Allow no empty line between fields (77f34c75f0436963feccf9a4f51f64f599d22ffb)

Let me know if you want to add any other modules to implement additional rules or otherwise change the configuration. It is OK that we are leaving out the `JavadocMethod` module for now, as well as to have those TODOs in the type comments.

Thanks again for this contribution, @hustodemon!

hustodemon:
Cool, thanks for merging!

PR #51, opened 2015-03-11
Fix: send kwargs properly to the netapi and adjust tests

### LINKED ISSUE #49: Clarify usage of kwargs
@renner can you clarify what the `kwargs` parameter in methods like `startCommand` is used for? As far as i can tell it only adds additional properties to the JSON but i don't know what i would use that for.

Initially i thought it would be to pass parameter to the remote function by name instead of by position but after a bit of research i found out that the `kwarg` property is used for that.

lucidd:
This fixes #49 and adjusts the test to make sure everything works as expected.

One thing to note here is that we pass the arguments all arguments (positional and named) as `json` strings. At first i thought that would be a problem but after some tests is looks like the server is fine with it and parses the strings to the type of the argument and uses them as expected.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/utils/ClientUtils.java
+++ b/src/main/java/com/suse/saltstack/netapi/utils/ClientUtils.java
@@ -56,17 +56,26 @@ public class ClientUtils {
     }
 
     /**
-     * Helper for constructing json object from kwargs and args.
+     * Helper for constructing json object with kwargs and args.
      *
      * @return JsonObject filled with kwargs and args.
      */
-    public static JsonObject makeJsonData(Map<String, String> kwargs, List<String> args) {
+    public static JsonObject makeJsonData(Map<String, String> props,
+            Map<String, String> kwargs, List<String> args) {
         final JsonObject json = new JsonObject();
 
-        if (kwargs != null) {
-            for (Map.Entry<String, String> kwEntry : kwargs.entrySet()) {
-                json.addProperty(kwEntry.getKey(), kwEntry.getValue());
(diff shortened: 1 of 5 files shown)


src/main/java/com/suse/saltstack/netapi/utils/ClientUtils.java
```diff
@@ -56,17 +56,25 @@ public static String streamToString(InputStream inputStream) {
     }
 
     /**
-     * Helper for constructing json object from kwargs and args.
+     * Helper for constructing json object with kwargs and args.
      *
      * @return JsonObject filled with kwargs and args.
      */
-    public static JsonObject makeJsonData(Map<String, String> kwargs, List<String> args) {
+    public static JsonObject makeJsonData(Map<String, String> props, Map<String, String> kwargs, List<String> args) {
```
isbm: Wrap here, perhaps?

renner: I actually hope we can merge #50 soon, and maybe this patch could be rebased on top afterwards. Then there is no way around satisfying checkstyle first.

lucidd: I agree i think it makes sense to merge #50 before all other PRs. I will rebase this as soon as #50 is done.

renner:
Ok, feel free to rebase this now, I just merged #50.

lucidd:
Rebased again ontop of the latest master and fixed all styling complains. :)

moio:
Well done @lucidd, merged!

PR #54, opened 2015-03-16
EAuth enhancements

lucidd:
This PR changes the eauth string constants to an enum and adds the missing values from the [docs](http://docs.saltstack.com/en/latest/ref/auth/all/). It also changes the default value for the login method without eauth parameter to `PAM` instead of `AUTO` since the [docs](http://docs.saltstack.com/en/latest/ref/auth/all/salt.auth.auto.html#module-salt.auth.auto) say `AUTO` is not intended for production use.


renner:
The patch looks good to me, very nice enhancement! I only wonder if `PAM` should really be the default? I know that `AUTO` is not intended for production use and `PAM` might be the easiest to set up, but from my perspective we could keep `AUTO` as the default.

lucidd:
**Update**: I forgot to change the `run` methods eauth parameter. That's now fixed.

In the current state having `AUTO` as the default value would make the `login` with 2 parameters practically useless for everything besides testing.

I think there are 2 options we could go with:

* Removing the login with 2 parameters completely.
* Putting the default AuthModule in the config so everybody can used it after configuration.

I would tend more towards removing it since its not that much work to pass eauth explicitly and to keep the API consistent we would also have to add 2 more functions for `run` to provide the same defaults.

renner:
I would then vote for removing the two parameter login method completely. Let's do that and not use a default `eauth` module anywhere.

lucidd:
Ok i removed it. Should be fine now. Thanks for the review.

moio:
Merging as per previous comments. Thanks everyone for cooperation!

PR #55, opened 2015-03-17
Implement keys endpoint with no parameters.

lucidd:
This PR implements the basic `/keys` endpoint with no parameters. There are 3 more endpoints related to key management which will be part of separate PRs once i figure out the details.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
@@ -4,6 +4,7 @@ import com.suse.saltstack.netapi.Constants;
 import com.suse.saltstack.netapi.client.impl.HttpClientConnectionFactory;
 import com.suse.saltstack.netapi.config.ClientConfig;
 import com.suse.saltstack.netapi.config.ProxySettings;
+import com.suse.saltstack.netapi.datatypes.Keys;
 import com.suse.saltstack.netapi.datatypes.cherrypy.Stats;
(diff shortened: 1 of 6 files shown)


lucidd:
@hustodemon thanks for the feedback i fixed the comment and naming issue. I went with acceptedMinions because it fits well with the rejectedMinions name.  As for the newline issue you pointed out on `keys_response.json` i couldn't a missing newline there neither locally nor on github. Can you clarify what you me you meant and look if its still a problem?

PS: make sure you comment on the PR instead of the commit in my repository so your comments are counted and don't disappear. :smile:

renner:
The patch looks good to me, thanks to @hustodemon for the review! It doesn't work for me though, I get response code `500` from the server with my current setup while the master log is show this:

    [WARNING ] Authentication failure of type "token" occurred for user saltdev.

I might be missing something, do I need to login as root in order to list the keys or how to setup authentication correctly? If root is needed we might want to mention that in the javadoc of these methods.

lucidd:
@renner From the docs it seems to require the `@wheel` permission: `These URLs wrap the functionality provided by the key wheel module functions.` After adding that it worked for me.

renner:
Ah thanks, you are right, works fine after adding `@wheel`. We might want to mention that in the javadoc as well maybe? Apart from that this patch can be merged.

lucidd:
Good idea i added it to the docs. I think we should do the same for the other functions too.

moio:
Merging as per previous comments. Thanks everyone for cooperation!

PR #56, opened 2015-03-17
Fix maven warnings

lucidd:
This makes the maven warnings go away and the encoding no longer dependent on the platform.


moio:
LGTM, thanks!

PR #58, opened 2015-05-01
Updates to support rest_cherrypy /events API.

bgaisford:
There's still a bit of work to do here.  I thought it best I create a PR and get some feedback from the other contributors before moving forward.  Still working through how best to create a unit test for this event stream.


renner:
Hello and thanks for contributing! This is a very nice effort overall, it will be super to include support for the `/events` API. I like this initial version, just tested it and it works great. Please feel free to move this forward with the goal of having a first version that can be merged.

Here is some questions that came up as I tried it out:

- I need to log in before I can consume the stream, do we want to mention that somewhere?
- How would I close the event stream from my application?
- How could this code be tested automatically? Please add some unit tests if possible!

pom.xml
```diff
@@ -56,6 +56,23 @@
...
+            <phase>package</phase>
+            <goals>
+              <goal>copy-dependencies</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${project.build.directory}/lib</outputDirectory>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
```
renner: Why was this added, do we really need it?

pom.xml
```diff
@@ -72,6 +89,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-all</artifactId>
+      <version>1.10.19</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
```
renner: This is not referenced anywhere in the code, right? I would guess that [wiremock] (http://wiremock.org/) should have everything that is needed to test this feature.

bgaisford:
@renner  Thanks for your feedback, I'm glad you like the new additions.  I'll add line notes to a couple of your queries above.  Regarding your main points:

* Agree, the Javadocs should be updated to indicate user must first login prior to subscribing/listening to the events stream
* Regarding closing the event stream, a new method `closeStream()` should be added to  `SaltStackEventStream`.  Once called all subscribed listeners should be notified via `eventStreamClosed()` and listeners cleaned up and steam closed.
* Yes, still working through the details of the unit test - seems to be more difficult than writing the original code :)

What do you think about the class names in general?  I'm thinking it may be best to remove Salt or SaltStack from my class additions.  I've gone back and forth on this in my mind on this.  In looking at the existing classes in the package, it would probably be more faithful to remove Salt/SaltStack.

pom.xml
bgaisford: As I was prototyping the unit test/s, I found [mockito](http://mockito.org/) to have better support for what I was trying to do.  I'll be sure to use [wiremock](http://wiremock.org/) first if it supports what is needed to implement the test.

pom.xml
bgaisford: As I was testing the client I found this useful for packaging the libraries needed to run the client standalone from the command line.

renner:
Sounds good, the caller should be able to close the stream as you suggest I guess. Testing this could indeed get a bit tricky, but nevertheless we should do it. Let me know if there is any problems that you need help with.

About the class names: Good point, we decided a while ago (in #24) to drop the *Salt*/*SaltStack* prefixes everywhere besides the main library class (`SaltStackClient`). Please remove those prefixes from your new classes as well, thanks!

pom.xml
renner: Aha I see. Feel free to continue working this way, but maybe remove this change from the patch?

bgaisford: Will do.

bgaisford:
@renner Completed coding updates based on your feedback.  Also did some additional general cleanup and finalized outstanding TODO items.  I think the code is pretty much where I want it unless you have further feedback.  I'll start on the unit test/s now.

bgaisford:
@renner I thought of one area you could certainly help me out.  The current code as per the PR doesn't attempt to deserialize the events into Java POJOs.  Perhaps the Jersey SSE implementation is returning well defined JSON messages that could be easily deserialized - I haven't checked yet.  Given your testing facilities there, perhaps you could generate a large body of diverse Salt server events so we can determine the level of difficulty in providing event deserialization support for the Java client.

bgaisford:
Not planning to work on this further until I get some feedback otherwise.  Cheers!

src/main/java/com/suse/saltstack/netapi/event/impl/JerseyServerSentEvents.java
```diff
@@ -0,0 +1,224 @@
...
+        WebTarget target = client.target(config.get(ClientConfig.URL) + "/events");
+        Invocation.Builder builder = target.request(new MediaType("text",
+                "event-stream"));
+        builder.header("X-Auth-Token", config.get(ClientConfig.TOKEN));
+
+        eventInput = builder.get(EventInput.class);
+    }
+
+    /**
+     * This method initiates the event processing loop.  It was added
```
moio: I guess the comment line was truncated for some reason :smile:

src/test/java/com/suse/saltstack/netapi/event/JerseyServerSentEventsTestCase.java
```diff
@@ -0,0 +1,258 @@
...
+
+        Client jerseyClient = ClientBuilder.newClient();
+        JerseyServerSentEvents jerseyServerSentEvents = new JerseyServerSentEvents(
+                client.getConfig(), jerseyClient);
+
+        EventCountClient eventCountClient = new EventCountClient(6);
+        jerseyServerSentEvents.addEventListener(eventCountClient);
+
+        jerseyServerSentEvents.processEvents();
+
```
moio: Trailing blank line here, please remove them.

renner:
@bgaisford: I will test your patch again, thanks for all those fixes and for adding tests! I might have some more remarks after testing (and @moio might add some more comments as well), but the overall scope of this PR is fine. Let's do the deserialization of events in another pull request after we merged this one.

moio:
Hey @bgaisford, thanks for this contribution!

I must say it's a pleasure to take the occasion of deep-diving into new tech while reviewing a PR, as it was the case with SSE here, :+1:!

Now a few questions:
* jersey-media-sse - can you explain why you picked this library and not others?
* `EventStream` - what about extending `AutoCloseable`?
* `EventStreamFactory` - while I appreciate code being extendable and I like your implementation, I am not sure if it is worth to add a Factory here since we only have one `EventStream`/`EventListener` implementation. The main reason is: does it make sense to add and maintain an infrastructure to support several libraries? In other words, will it ever make sense to add and maintain more than one backend here? What about consolidating on Jersey and removing this Factory, the separate `EventListener` interface, `EventStreamType` and `EventStream`?
* `JerseyServerSentEvents.processEvents()` - this method seems public for testing reasons only, correct? In that case I would suggest to add a Javadoc comment to tell clients that method should not be used directly.
* `EventAdapter` - what is this for?
* `JerseyServerSentEventsTestCase` - we named all test classes `*Test` and not `*TestCase`, please adapt this name accordingly.

About generating diverse Salt server events - I agree with @renner, this is not needed for this PR.

Overall those are minor points, I really like your contribution, thank you!

bgaisford:
@moio @renner Thanks for the excellent feedback, it's very much appreciated.  I also like this feature and can envision uses for it as well as much more work that needs to be done:

1. Event deserialization (next PR)
2. Event versioning - SSE has very generic wrappers around the underlying Salt/rest_cherrypy events subsystem.  Over time the underlying event message contents may change.  How should versioning be handled?
3. Event filters would be a nice addition - only subscribe to certain types of events.

Addressing your points above (in order):    

* To be honest, I didn't do a thorough search of alternative Java SSE implementations.  The GlassFish/Jersey version was the first one I found.  Given that GlassFish is a reference platform a lot of times, it seems like a good initial candidate library for this feature.
* Agreed, will do.
* The factory/facade pattern is important here.  Consumers of this API need isolation from possible backend changes.  API developers need latitude to change the backend implementation without breaking existing client implementations.  I agree with removal of `EventStreamType` and associated logic.  Only one backend implementation will be supported.
* Agreed, will do.
* I'll remove `EventAdapter `.  When I first started this effort I thought the `EventListener ` interface may containing many more methods than it did.  I added an adapter so a client only interested in a small subset of the `EventListener` methods could opt out of providing implementations for all of them.
* Agreed, will do.

Thanks again for all your feedback and encouragement, this was a fun feature to implement. :)  I'll commit these latest updates promptly.

**Note:** While looking at the project forks on GitHub yesterday I noticed that @mohamedazouz has a branch to add events.  I included him here in this PR comment in hopes to avoid duplication of effort.

cmorar:
@bgaisford  and @renner  I did some implementation for parsing events on my branch. And also I have a  pretty ugly test implementation which instead of Jersey uses plain Apache HTTPClient - uses lonh polling . Unfortunately I don't have the time needed to get this done but if you guys find it useful fell free to use it .. at least as a starting point for events parsing

bgaisford:
@cmorar Thanks for the heads up, I'll definitely take a look at your work to date.

renner:
Hey @bgaisford, I merged your patch after I tested it again. Works great so far, seeing all the events coming in real time. Please have a look at the commits I added on top. I discussed with @moio about the `EventStreamFactory` (and interface) and we both agreed on "don't add complexity until it's needed".

The point is: whenever we want to change the underlying implementation we will just do that. We won't break the API since the class is just called `EventStream` now and we further won't have two different implementations in parallel that you can switch between.

Two other remarks that you should consider when doing further contributions:

- Some of your commit messages are quite long, please note that commit messages should better follow certain guidelines, see for example [this article] (http://chris.beams.io/posts/git-commit/).
- Fixing the typo ("Required persmissions") is a very valid fix (and I merged it), but it doesn't belong to this patch. Better do such a fix as a separate PR or at least in a separate commit.

Apart from that: thanks again for your contribution and please feel free to submit any further pull requests!

bgaisford:
Hey @renner, the updates look great.  Thanks for the additional pointers as well.  Cheers!

PR #59, opened 2015-05-03
Add job start time to Job class

popstr:
Adds start time to Job class.

Salt JobIds are made from the date and time when the job was created (example from today - May 3, 2015: 20150503195439947382). Unfortunately, Salt Master does not use UTC when constructing these Id's, meaning the master's current time zone is used which is unknown to the minion and so it gets lost when translating the JobId into a StartTime.

One could argue that it would be easier to just parse the JobId instead of the strange date format that cherrypy returns, but I don't think we should rely on JobId's having this format forever.

It would be much better if cherrypy returned a timestamp instead. That would mean that the minion can translate the date and time into it's own time zone. Parsing would also be easier and more reliable.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/datatypes/Job.java
+++ b/src/main/java/com/suse/saltstack/netapi/datatypes/Job.java
@@ -1,14 +1,21 @@
 package com.suse.saltstack.netapi.datatypes;
 
+import com.google.gson.annotations.JsonAdapter;
 import com.google.gson.annotations.SerializedName;
+import com.suse.saltstack.netapi.parser.JsonParser;
 
-// TODO - handle dates too (they are represented in really
-// exotic format: (2015, Mar 04 19:28:29.724698))
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Locale;
 
 /**
  * Representation of a previously run job.
  */
 public class Job {
+
+    // StartTime example from API: "2015, Mar 04 19:28:29.724698"
+    public static final SimpleDateFormat START_TIME_FORMAT =
+            new SimpleDateFormat("yyyy, MMM dd HH:mm:ss.SSS", Locale.US);
 
     @SerializedName("Function")
     private String function;
@@ -24,6 +31,13 @@ public class Job {
 
     @SerializedName("Arguments")
     private Arguments arguments;
+
+    /**
+     * Please note that start time will be in salt-master time zone
+     */
+    @SerializedName("StartTime")
+    @JsonAdapter(JsonParser.JobStartTimeJsonAdapter.class)
+    private Date startTime;
 
     public String getFunction() {
         return function;
@@ -43,5 +57,9 @@ public class Job {
 
     public Arguments getArguments() {
         return arguments;
+    }
+
+    public Date getStartTime() {
+        return startTime;
     }
 }
--- a/src/main/java/com/suse/saltstack/netapi/exception/ParsingException.java
+++ b/src/main/java/com/suse/saltstack/netapi/exception/ParsingException.java
@@ -1,9 +1,11 @@
 package com.suse.saltstack.netapi.exception;
 
+import java.io.IOException;
+
 /**
  * Exception to be thrown in case of problems parsing service responses.
  */
-public class ParsingException extends SaltStackException {
+public class ParsingException extends IOException {
 
     /**
      * Constructor expecting a custom cause.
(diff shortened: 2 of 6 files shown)


renner:
Hello Mikael, thanks a lot for contributing! I agree that we should not rely on the format of the job id, we need to parse the start time as we get it from the API. Therefore I am more or less fine with the proposed patch, here are just two remarks:

- There is a bunch of deserializers already in [`JsonParser.java`] (https://github.com/SUSE/saltstack-netapi-client-java/blob/master/src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java#L86), could we add a `StartTimeDeserializer` there instead of the proposed `TypeAdapter`?
- Can we cover this by a unit test, either with a new test in `JsonParserTest.java` or in `SaltStackClientTest.java`?

I also opened an [issue about this] (https://github.com/saltstack/salt/issues/23349).

src/main/java/com/suse/saltstack/netapi/datatypes/Job.java
```diff
@@ -1,15 +1,27 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Locale;
 
 /**
  * Representation of a previously run job.
  */
 public class Job {
 
+    // StartDate example from API: "2015, Mar 04 19:28:29.724698"
```
renner: We should call it `StartTime` everywhere so we won't confuse ourselves.

renner:
Please note that checkstyle failed on your current patch, please don't use `.*` imports:

    Using the '.*' form of import should be avoided - java.io.*.

moio:
Hey @popstr, thanks for this nice contribution!

This was really needed, and generally I like the implementation. Only one small remark from my side: we should not swallow `Exception`s if the date is not parseable, we should throw an unchecked `Exception` instead.

Google's example (`PointAdapter`) throws `NumberFormatException` and we should do something similar I think:

https://google-gson.googlecode.com/svn/trunk/gson/docs/javadocs/com/google/gson/TypeAdapter.html#read%28com.google.gson.stream.JsonReader%29
https://docs.oracle.com/javase/7/docs/api/java/lang/Integer.html#parseInt%28java.lang.String%29

Do you agree?

Thanks again, good job!

popstr:
@renner Is it possible to add a StartTimeDeserializer that handles dates for a specific field only? In the constructor for JsonParser the Deserializers are initialized as follows:

    this.gson = new GsonBuilder()
                .registerTypeAdapter(Date.class, new SaltStackDateDeserializer())
                .registerTypeAdapter(Stats.class, new StatsDeserializer())
                .registerTypeAdapter(Arguments.class, new ArgumentsDeserializer())
                .create();

As you can see, all Date fields are handled by the same Deserializer.

popstr:
@moio Good point! There is an Exception class called ParsingException (in com.suse.saltstack.netapi.exception) that would be appropriate here. The problem is that it is a checked Exception so in that case we have to declare the read method in the TypeAdapter to throw ParsingException, which we probably want to avoid.
There are 3 alternatives here;
1. Create another exception class extending RuntimeException (could cause confusion to have 2 Exceptions for the same purpose)
2. Make ParsingException extend RuntimeException instead of Exception
3. Make ParsingException extend IOException instead, which will work because the read method is already throwing IOException. Since a parsing error really is an Input/Output error, this makes sense to me.
I think option 3 is the most logical and least disruptive alternative. As far as I can see, ParsingException is not used anywhere right now, so this change would not create any side effects.

moio:
@popstr, fine in making `ParsingException` an `IOException` from my POV.

As soon as you deal with other comments we will merge your PR! Thanks!

renner:
@popstr Apparently that is not possible: looking at the [available annotations] (https://google-gson.googlecode.com/svn/trunk/gson/docs/javadocs/com/google/gson/annotations/package-summary.html) it seems that you can only specify `TypeAdapter` classes to be used for certain fields, as you are doing it in your patch. Reading on about the [JsonDeserializer interface] (https://google-gson.googlecode.com/svn/trunk/gson/docs/javadocs/com/google/gson/JsonDeserializer.html) they actually advise users to use a `TypeAdapter` instead:

    New applications should prefer TypeAdapter, whose streaming API is more efficient than this interface's tree API.

So let's do that and keep your `TypeAdapter` as it is. Further (in a separate pull request) all existing `JsonDeserializer` implementations could be converted to `TypeAdapter` as well.

popstr:
Added tests for StartTime parsing and the TypeAdapter now throws a ParsingException when there is a parse error.

renner:
Hey @popstr, please note that one of your tests is [failing] (https://travis-ci.org/SUSE/saltstack-netapi-client-java/jobs/61634856).

popstr:
@renner That's weird. The test runs fine on my machine. What if you check out my branch and run the tests on your machine?

popstr:
It is a time zone issue. The test runs fine on my machine, but the Travis server is in another time zone. I've pushed an update that verifies the date without using timezone.

renner:
Thanks for the fix, I merged your patch. The testsuite needs to succeed on the travis host, of course. Thanks a lot for this contribution and all your efforts, feel free to go on contributing!

PR #61, opened 2015-05-05
Add events

mohamedazouz:
add /events


renner:
Hello and thanks for contributing! Please note that there is already some ongoing effort to implement support for the `/events` endpoint, see #58. I would therefore advise you to rather focus on something else if you would like to contribute to this library.

Apart from that please note that `/events` is not an endpoint that simply returns a list of events to be parsed into objects, your current patch would therefore never return any events. What we rather get is an HTTP stream of the salt master event bus and that stream is formatted per the Server Sent Events (SSE) spec. This allows us to receive events from the salt master in real time which has been implemented in #58.

Sorry that I am closing this, hope to see more contributions from you!

PR #62, opened 2015-05-11
Implement call to /hook

mateiw:
Implement methods in SaltStackClient to call /hook.
The event data is passed as a string. It's not the prettiest way but this offers more flexibility to the client of this method. Another options would have been to pass it as a Gson JsonElement but that would force the caller to use Gson.


renner:
Hello @mateiw, thanks a lot for contributing! Your patch looks quite good to me, it is OK from my perspective to pass the event data as a string, at least in the first version. I will give it a try and then might come back to you with some more feedback.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -457,4 +457,30 @@ public Keys call() throws SaltStackException {
     public EventStream events() {
         return new EventStream(config);
     }
+
+    /**
+     *
+     * @param eventTag
+     * @param eventData
+     * @return
+     * @throws SaltStackException
```
renner: Please complete the javadoc: put a description of the method, parameters, return value and when an exception is thrown.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -457,4 +457,30 @@ public Keys call() throws SaltStackException {
...
+     */
+    public boolean hook(String eventTag, String eventData) throws SaltStackException {
+        String tag = eventTag != null ? eventTag : "";
+        Map<String, Object> result = connectionFactory
+                .create("/hook/" + tag, JsonParser.MAP, config)
+                .getResult(eventData);
+        return Boolean.TRUE.equals(result.get("success"));
+    }
+
+    public Future<Boolean> hookAsync(final String eventTag, final String eventData) {
```
renner: Please put a javadoc comment here as well.

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -457,4 +457,30 @@ public Keys call() throws SaltStackException {
...
+        Callable<Boolean> callable = new Callable<Boolean>() {
+            @Override
+            public Boolean call() throws Exception {
+                return hook(eventTag, eventData);
+            }
+        };
+        return executor.submit(callable);
+    }
+
```
renner: Please remove this trailing empty line, thanks.

renner:
I tested your patch successfully after setting up a reactor on the salt master in my test environment and figuring out how to access the data inside the reactor. We will merge it, but please consider the following remarks first:

- Could you please rebase the patch on top of the current master branch? There is some conflicts now after we merged another patch.
- @moio mentioned that the method name `hook()`can be misleading since actually what you are doing is triggering an event. His proposal is therefore to rename it to `sendEvent()` and I would agree with that. Other options would be `triggerEvent()` or `fireEvent()`, feel free to choose one of those.

About the naming please note that in a [previous PR discussion] (https://github.com/SUSE/saltstack-netapi-client-java/pull/15#issuecomment-74485966) we decided to use the API endpoint names, but only as long as this actually makes sense. Thanks for your efforts so far!

src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
```diff
@@ -457,4 +457,40 @@ public Keys call() throws SaltStackException {
...
+                .getResult(eventData);
+        return Boolean.TRUE.equals(result.get("success"));
+    }
+
+    /**
+     * Asynchronously trigger an event in Salt with the specified tag and data.
+     *
+     * POST /hook
+     *
+     * @return Future containing a boolean value indicating the success of failure of the event triggering .
```
renner: Sorry to bother you again with this, but please describe the parameters here as well.

mateiw:
I've changed the name. Indeed it makes more sense for this method to be called `sendEvent()`. Fixed the javadoc too and rebased.

moio:
Merged, thank you very much for this contribution Matei, see you at the next pull request!

:+1: :smiley:

PR #65, opened 2015-05-14
List minions

emre-aydin:
Implemented getMinions method for "GET /minions" endpoint


src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
```diff
@@ -32,25 +21,33 @@
...
-import static com.github.tomakehurst.wiremock.client.WireMock.any;
+import static com.github.tomakehurst.wiremock.client.WireMock.verify;
+import static com.suse.saltstack.netapi.AuthModule.AUTO;
+import static com.suse.saltstack.netapi.AuthModule.PAM;
+import static com.suse.saltstack.netapi.config.ClientConfig.SOCKET_TIMEOUT;
 import static org.hamcrest.CoreMatchers.containsString;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
```
moio: It looks like your IDE has automatically rearranged import statements here. It is not a big deal, but in general I would suggest to minimize changes introduced by a patch by just adding the new classes you need. In case a reorganization is really needed, please provide it as a separate patch or even PR.

Please revert changes here and only add lines for classes that are needed, thank you!

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
```diff
@@ -129,7 +128,7 @@ public void testLoginFailure() throws Exception {
                 .withHeader("Accept", equalTo("application/json"))
                 .withHeader("Content-Type", equalTo("application/json"))
                 .willReturn(aResponse()
-                .withStatus(HttpURLConnection.HTTP_UNAUTHORIZED)));
+                        .withStatus(HttpURLConnection.HTTP_UNAUTHORIZED)));
```
moio: Whitespace was added here, and it looks unrelated to your PR - please try to limit your patch to the minimum change needed. Revert this change please.

moio:
Hey @emre-aydin, thanks for your contribution! Please check out Travis CI, as it seems there are some checkstyle errors. I noted a couple of small points directly in code - nothing really substantial but have a look.

I am talking with colleagues today about other aspects of your PR - other comments will be posted soon.

Have a nice day!

renner:
Hey @emre-aydin, when looking at the `get_minions_response.json` file I understand why you are returning a `List<Map<String, Object>>` instead of a list of `Minion` objects. I am fine with that for the scope of this pull request, but please mention in the javadoc of the new method that various data will be returned and ideally even add a link to some reference listing the possible keys in there. If there is no reference to be found anywhere throughout the salt documentation and source code, the json file itself could serve as a reference for now I guess.

Further your unit test could actually verify some more of the test data in the json file, at least I would suggest to have all the different data structures in there covered. You could check also the size of the returned `Map` contents to somehow verify completeness, since verifying the complete dataset would be too much effort indeed.

One more thing that you could add to this PR is support for the `GET /minions/(mid)` endpoint. i.e. you could add another method to the library that takes a minion ID as parameter and returns the minion details, like `getMinionDetails(String minion)`. What do you think?

emre-aydin:
Hey all. There's some documentation on listing all grains [here](http://docs.saltstack.com/en/latest/topics/targeting/grains.html#listing-grains) but I think there's none which shows them as a list, so better use the json file itself as a reference for now.

Totally agree with tests, and nice idea to add support for `GET /minions/(mid)`, will add it.

emre-aydin:
> the json file itself could serve as a reference for now I guess

Not sure how to do this, is there a way to reference this file in javadocs?

renner:
I like the patch now, @emre-aydin. In the javadoc I would recommend to just put the link to the official [grains documentation] (http://docs.saltstack.com/en/latest/topics/targeting/grains.html), where it is also described how to list all available grains as you mentioned before.

renner:
I merged it, thanks for your efforts and for fixing everything as requested by the reviewers! :+1: Please feel free to continue contributing in terms of submitting further pull requests.

PR #66, opened 2015-05-15
Cleaning up Tests 

lucidd:
This cleans up / simplifies the tests and addresses all the points of #43.

### DIFF
--- a/src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
+++ b/src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
@@ -189,13 +189,41 @@ public class SaltStackClientTest {
             }
         };
 
-
-        client.run("user", "pass", PAM, "local", "*", "pkg.install", args, kwargs);
+        Map<String, Object> retvals =
+                client.run("user", "pass", PAM, "local", "*", "pkg.install", args, kwargs);
 
         verify(1, postRequestedFor(urlEqualTo("/run"))
                 .withHeader("Accept", equalTo("application/json"))
                 .withHeader("Content-Type", equalTo("application/json"))
                 .withRequestBody(equalToJson(JSON_RUN_REQUEST)));
+
+        Map<String, Map<String, String>> expected =
+                new LinkedHashMap<String, Map<String, String>>(){
+            {
+                put("i3-wm", new LinkedHashMap<String, String>(){
+                    {
+                        put("new", "4.10.1-1");
+                        put("old", "");
+                    }
+                });
+                put("i3lock", new LinkedHashMap<String, String>(){
+                    {
+                        put("new", "2.6-1");
+                        put("old", "");
+                    }
+                });
+                put("i3status", new LinkedHashMap<String, String>(){
+                    {
+                        put("new", "2.9-1");
+                        put("old", "");
+                    }
+                });
+            }
+        };
+
+        assertNotNull(retvals);
+        assertTrue(retvals.containsKey("minion-1"));
+        assertEquals(retvals.get("minion-1"), expected);
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


moio:
I agree, we can merge this. From a purely stylistic POV, I usually write `Collection`s "initializers" like this:

    List<String> myList = new LinkedList<String>() {{
        add("a");
        add("b");
        add("c");
    }};

This kind of breaks at least my automatic code formatter, as it would put newlines between curly braces, but looks clearer in its intention to me. Feel free to change your code accordingly or not!

lucidd:
@moio I agree this would look much better and more concise but it conflicts with the current checkstyle rules in multiple ways. A checkstyle compliant version would look like this:

```java
Map<String, String> kwargs = new LinkedHashMap<String, String>() { {
        put("refresh", "true");
        put("sysupgrade", "false");
    } };
```
Its still better then the current version despite the extra indentation and whitespace but it gets really awkward when nesting maps.
I'm not sure if we can configure checkstyle in a way that it accepts your version while still enforcing those rules in other places where they make sense.
Also the <> operator does not work for anonymous inner classes so we are stuck with the longer version.

renner:
Ok, let's merge it. I like @moio's version best while I'm not sure if I like the checkstyle compliant version better than the style in the commit. Feel free to do as you prefer. And could you please rebase against the current master, there is some conflicts after the merge of #65.

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
```diff
@@ -212,14 +241,43 @@ public void testRunRequestAsync() throws Exception {
...
+        Future<Map<String, Object>> future = client.runAsync("user", "pass",
+                PAM, "local", "*", "pkg.install", args, kwargs);
+        Map<String, Object> retvals = future.get();
 
         verify(1, postRequestedFor(urlEqualTo("/run"))
                 .withHeader("Accept", equalTo("application/json"))
                 .withHeader("Content-Type", equalTo("application/json"))
                 .withRequestBody(equalToJson(JSON_RUN_REQUEST)));
+
+
```
renner: Maybe remove one of those double empty lines here.

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
```diff
@@ -335,15 +359,44 @@ public void testQueryJobResult() throws Exception {
 
         Map<String, Object> retvals = client.getJobResult("some-job-id");
 
+        verify(1, getRequestedFor(urlEqualTo("/jobs/some-job-id"))
+                .withHeader("Accept", equalTo("application/json")));
+
+
```
renner: Same as above, maybe remove one empty line?

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
```diff
@@ -335,15 +359,44 @@ public void testQueryJobResult() throws Exception {
...
+                        put("old", "");
+                    }
+                });
+                put("i3lock", new LinkedHashMap<String, String>(){
+                    {
+                        put("new", "2.6-1");
+                        put("old", "");
+                    }
+                });
+                put("i3status", new LinkedHashMap<String, String>(){
```
renner: If we keep this collection initialization style, I would personally put a blank between all the `()` and the `{`. Might be only a personal preference though.

moio:
@lucidd, you are right. I just double checked and it seems there is no way for Checkstyle to digest "double brace initialization" at the moment.
Solutions could be:
 - patch Checkstyle! (in a totally different effort)
 - disable indentation checking. This is what we do in Spacewalk;
 - keep the very bulky Checkstyle-approved style or avoid double brace initialization altogether.

Your choice.

lucidd:
@renner I rebased it and removed the double new lines. I also adjusted the new tests from #65 to this patch. @moio I will leave the changes to the double brace ininitialization out of this patch and submit a new one once i tried out a few of those options. If its ok with you guys i will merge this as soon as travis is ok with it.

renner:
Fine with me, feel free to merge :+1:

Issue #64, opened 2015-05-12
Logout could return a boolean instead of a string result


renner:
The current `logout()` method returns a `Result<String>`, which in my test environment weirdly even says "Welcome". While posting to the `/logout` endpoint should actually return ["Your token has been cleared"] (https://github.com/saltstack/salt/blob/develop/salt/netapi/rest_cherrypy/app.py#L1481), I guess we could change the return type of the `logout()` to boolean (in case the string matches the above).

renner:
I found out that `logout()` is actually broken since we don't POST: passing `null` as data will create a GET request, which is also the reason why "Welcome" is returned. I just fixed this in a branch with dcc30c25765dbd89bbe2ab05553676f6e2403ffe. I will further change the method signature of `logout()` to return a boolean and then make a pull request.

renner:
Fixed with #71.

PR #71, opened 2015-05-26
Fix logout to actually perform a POST request

renner:
This patch should fix the `logout()` and `logoutAsync()` methods to actually perform a POST request. Recently we accidentally sent a GET request here since the `data` passed in to `create()` was `null`. The proposed patch is just a quick fix, we might want to reconsider the behaviour of the `create()` method that does actually create a GET request whenever the given data is `null`.

Further this pull request adds missing unit tests for `logout()` and `logoutAsync()` and the example response (`logout_response.json`) has been fixed.

This fixes issue #64.


lucidd:
LGTM

moio:
Feel free to merge this.

renner:
Thanks for your reviews, merged.

PR #74, opened 2015-05-31
Replace SaltStackDateDeserializer with DateAdapter

hustodemon:
TypeAdapters are stream-oriented and more memory effecient than
(tree-based) Deserializers.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
+++ b/src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
@@ -72,7 +72,7 @@ public class JsonParser<T> {
     public JsonParser(TypeToken<T> type) {
         this.type = type;
         this.gson = new GsonBuilder()
-                .registerTypeAdapter(Date.class, new SaltStackDateDeserializer())
+                .registerTypeAdapter(Date.class, new DateAdapter().nullSafe())
                 .registerTypeAdapter(Stats.class, new StatsDeserializer())
                 .registerTypeAdapter(Arguments.class, new ArgumentsDeserializer())
                 .create();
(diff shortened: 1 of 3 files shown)


renner:
Makes sense and patch looks good to me as well. Are you going to convert also the other implementations of `JsonDeserializer` to `TypeAdapter` (there is two more)? We can do these separately I guess, created issue #75 for tracking that.

hustodemon:
sure. conversion of the other deserializers in in progress, i'd push it separately.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -92,18 +92,22 @@ public T parse(InputStream inputStream) {
...
-                throws JsonParseException {
+        public void write(JsonWriter jsonWriter, Date date) throws IOException {
+            throw new UnsupportedOperationException("Writing JSON not supported.");
+        }
+
+        @Override
+        public Date read(JsonReader jsonReader) throws IOException {
             try {
-                double dateMiliSecs = jsonElement.getAsDouble() * 1000;
+                double dateMiliSecs = jsonReader.nextDouble() * 1000;
```
renner: Very minor nitpick: milliseconds is written with double "l". I would suggest to use `dateMilliseconds` as the name for this variable (milliseconds is only one word).

hustodemon: That's right, I didn't know that.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -72,7 +72,7 @@
     public JsonParser(TypeToken<T> type) {
         this.type = type;
         this.gson = new GsonBuilder()
-                .registerTypeAdapter(Date.class, new SaltStackDateDeserializer())
+                .registerTypeAdapter(Date.class, new DateAdapter().nullSafe())
```
renner: As already discussed: I'm not sure if we really want this to be null safe, it would mean we accept **any** missing dates in the returned JSON, right? I would personally prefer to fail if any expected date is missing. This way we might even uncover bugs in salt-api while we can still decide to accept that a date is missing per field once it actually occurs.

hustodemon: Sure, let's not change the functionality. I'll push a new version in a moment.

hustodemon: After closer investigating I discovered that the previous approach with `JsonDeserializer` is null-safe by default (GSON uses `TreeTypeAdapter` by defaut, which handles null cases).

That means `.nullSafe()` call is needed to preserve the existing functionality. I suggest merging the patch as is (I can maybe remove tests I added in this patch, although I think they are useful as they document the current behavior). Handling required/non-required fields is out of the scope of this patch.

renner: Ok, fine with me. I will then just merge the patch as it is and open an issue about handling optional fields.

hustodemon: Sure. Thanks!

PR #76, opened 2015-06-09
Add information for maven central publish

lucidd:
This is still work in progress but here are the initial changes to comply with the maven central publishing process.


renner:
Thanks @lucidd, I merged it. I hope there was no specific reason to use version 1.5 of the gpg plugin, increased it to 1.6 with a60420227f6d0a74d4d0870489af6b27649a2324.

lucidd:
@renner no there was no reason just didn't know that there was a newer version.

Issue #77, opened 2015-06-10
DateAdapter should not be null-safe per default


renner:
The current generically used [`DateAdapter`] (https://github.com/SUSE/saltstack-netapi-client-java/blob/master/src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java#L98) is *null-safe* per default: whenever an expected date is missing in the returned JSON the library would just silently accept that. Instead we should probably fail per default and rather support the declaration of optional fields allowing us to accept missing values for selected date fields that may or may not be returned by the API.

PR #78, opened 2015-06-11
Add parser for Optional types to avoid the null madness

lucidd:
This adds a parser for `Optional<T>` for all types `T` we can already parse.
The main motivation behind this is to banish `null` as much as possible in favor of the much saver abstraction `Optional<T>` which denotes a possibly absent value. This addresses concerns raised in #77


lucidd:
I just rebased this to the latest master. I would say we merge this now as is even if we are not able to completely enforce that non optional values are never null its still valuable to mark `Optional` values as such and have them parsed accordingly.

renner:
We can merge this if you need it, choose yourself if you want to merge it from here or include the commits with #109.

src/test/java/com/suse/saltstack/netapi/parser/JsonParserTest.java
```diff
@@ -202,6 +206,41 @@ public void testSaltStackJobsWithKwargsParser() throws Exception {
...
+        List<Optional<Integer>> expected = new LinkedList<>();
+        expected.add(Optional.of(1));
+        expected.add(Optional.of(2));
+        expected.add(Optional.of(3));
+        expected.add(Optional.empty());
+        expected.add(Optional.of(5));
+        assertEquals(expected, result.maybeInts);
+    }
+
+
```
renner: Let's put only one empty line between methods.

src/test/java/com/suse/saltstack/netapi/parser/OptionalTest.java
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Helper Type to test the Optional Parser
+ */
+public class OptionalTest {
+
+    public Optional<String> nullString = Optional.empty();
+    public Optional<String> valueString = Optional.empty();
+    public Optional<String> absentString = Optional.empty();
+    public List<Optional<Integer>> maybeInts;
+
```
renner: Let's get also rid of this one.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -127,6 +131,45 @@ public Date read(JsonReader jsonReader) throws IOException {
...
+                    if (in.peek() == JsonToken.NULL) {
+                        in.nextNull();
+                        return Optional.empty();
+                    } else {
+                        A value = innerAdapter.read(in);
+                        return Optional.of(value);
+                    }
+                }
+
+                public void write(JsonWriter out, Optional<A> optional) throws IOException {
```
renner: Aren't `read()` and `write()` overrides from `TapeAdapter`? Let's put the `@Override` annotation?

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -127,6 +131,45 @@ public Date read(JsonReader jsonReader) throws IOException {
...
+                    innerAdapter.write(out, optional.orElse(null));
+                }
+            };
+        }
+
```
renner: Please remove this one as well.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -127,6 +131,45 @@ public Date read(JsonReader jsonReader) throws IOException {
     }
 
     /**
+     * TypeAdaptorFactory creating TypeAdapters for Optional
+     */
+    private class OptionalTypeAdapterFactory implements TypeAdapterFactory {
+
+        @SuppressWarnings("unchecked")
+        public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> typeToken) {
```
renner: Public methods should generally have javadoc, consider to add some documentation.

lucidd: create is overridden from `TypeAdapterFactory` so its already documented there i put the `@Override` on it to make it clear.

lucidd:
I would prefer merging this PR first i will rebase #109 accordingly.

PR #85, opened 2015-07-08
Preparations for releasing to maven central

### LINKED ISSUE #72: Make jar file available via maven central
We should make the library jar file available via [maven central] (http://maven.apache.org/guides/mini/guide-central-repository-upload.html) so it will be easier for anyone to include it with their project via pom.xml.

renner:
This is the remaining changes to pom.xml needed to deploy artifacts into maven central (via [OSSRH staging repositories] (http://central.sonatype.org/pages/releasing-the-deployment.html#login-into-ossrh)). This should help us to close #72 with the next release.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,6 +73,14 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
         <version>2.10.1</version>
+        <executions>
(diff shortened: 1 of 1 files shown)


renner:
Merging it. Note that we are not adding a snapshot repository location for now since we are currently not planning to release snapshot versions.

PR #86, opened 2015-07-22
Move EventStream implementation from SSE to WebSocket

ncounter:
Hi, this PR contains a WebSocket implementation in EventStream: everything is based on javax.websocket, and is intended to change only technology, not the interface. Because of this, the code preserve basically the same construction, where possible.
Most of these changes are annotations and methods of a WebSocket implementation, and removal of Executors and Callable objects because not needed anymore.
A point to mention is the exceptions handling: to keep safe the existing interface in some cases exception couldn't be thrown but just a stack trace printed, and in EventListener class an error callback method could also be implemented.
In the tests part a glassfish.tyrus server is started and it uses a websocket endpoint from where events are sent back to the WebSocket EventStream.


dmacvicar:
Hey, thanks for this PR!

It comes at the right time (we were just looking into this).

I have never been able to setup the websocket part on the server side successfully (only tried SSE). Did you had to do anything else appart of installing ws4py?

When running a simple test program:

```java
package com.suse.saltstack.netapi;

import com.suse.saltstack.netapi.client.SaltStackClient;
import com.suse.saltstack.netapi.config.ProxySettings;
import com.suse.saltstack.netapi.datatypes.Token;
import com.suse.saltstack.netapi.event.EventListener;
import com.suse.saltstack.netapi.event.EventStream;
import com.suse.saltstack.netapi.exception.SaltStackException;

import java.net.URI;

public class FooTest {

    public static void main(String[] args) throws InterruptedException {
        SaltStackClient client = new SaltStackClient(URI.create("http://foo.suse.de:8765"));
        client.setProxy(new ProxySettings("someproxy.suse.de", 3128, "****", "****"));
        try {
            Token token = client.login("admin", "admin", AuthModule.AUTO);
            System.out.println(token.getToken());
            EventStream stream = client.events();

            try {
                stream.addEventListener(new EventListener() {
                    @Override
                    public void notify(String s) {
                        System.out.println("->" + s);
                    }

                    @Override
                    public void eventStreamClosed() {
                        System.out.println("-> CLOSED!!!!");
                    }
                });
            } catch (RuntimeException e) {
                System.err.println("RuntimeException exception: " + e.getMessage());
            }

            System.out.println("-- Sleep --");
            Thread.sleep(100000);
            stream.close();

        } catch (SaltStackException ex) {
            System.err.println("URISyntaxException exception: " + ex.getMessage());
        }
        System.out.println("-- End of Main --");
    }
}
```

Using as implementation:

```xml
<dependency>
  <groupId>org.glassfish.tyrus</groupId>
  <artifactId>tyrus-container-jdk-client</artifactId>
  <version>1.9</version>
</dependency>
```

Against our server (installed ws4py) I get:

```
Jul 22 23:53:16 hoag salt-api[8072]: Process Process-1:6:
Jul 22 23:53:16 hoag salt-api[8072]: Traceback (most recent call last):
Jul 22 23:53:16 hoag salt-api[8072]: File "/usr/lib64/python2.7/multiprocessing/process.py", line 258, in _bootstrap
Jul 22 23:53:16 hoag salt-api[8072]: self.run()
Jul 22 23:53:16 hoag salt-api[8072]: File "/usr/lib64/python2.7/multiprocessing/process.py", line 114, in run
Jul 22 23:53:16 hoag salt-api[8072]: self._target(*self._args, **self._kwargs)
Jul 22 23:53:16 hoag salt-api[8072]: File "/usr/lib/python2.7/site-packages/salt/netapi/rest_cherrypy/app.py", line 1973, in event_stream
Jul 22 23:53:16 hoag salt-api[8072]: json.dumps(data)), False)
Jul 22 23:53:16 hoag salt-api[8072]: File "/usr/lib/python2.7/site-packages/ws4py/websocket.py", line 261, in send
Jul 22 23:53:16 hoag salt-api[8072]: self._write(m)
Jul 22 23:53:16 hoag salt-api[8072]: File "/usr/lib/python2.7/site-packages/ws4py/websocket.py", line 243, in _write
Jul 22 23:53:16 hoag salt-api[8072]: self.sock.sendall(b)
Jul 22 23:53:16 hoag salt-api[8072]: error: [Errno 104] Connection reset by peer
Jul 22 23:53:16 hoag systemd-journal[5686]: Forwarding to syslog missed 2 messages.
```

And I see the stream immediately closed.

dmacvicar:
Nevermind. I see the events now. It seems related to the proxy and that the stream is closed very quickly when no events come for a few second.

ncounter:
Hi, thanks for your interests.
You're right, nothing to do more than install ws4py. Fine that you see the events.
WebSocket <code>setDefaultMaxSessionIdleTimeout</code> in EventStream is setted to the <code>com.suse.saltstack.netapi.config.ClientConfig.SOCKET_TIMEOUT</code> value. May the stream is closing because no events come in the default 10 seconds, and if you set this to <code>0</code> it should stay open (unless the proxy does something on its own).

src/test/java/com/suse/saltstack/netapi/event/TyrusWebSocketEventsTest.java
```diff
@@ -0,0 +1,272 @@
...
+        @Override
+        public void eventStreamClosed() {
+        }
+    }
+
+    /**
+     * Event listener client used for testing.
+     */
+    private class EventStreamClosedClient implements EventListener {
+        private EventStream eventStream;
```
renner: This field seems to be not used anywhere, I guess it could be removed from here as well as from the constructor?

ncounter: You're right, i moved the Assert but didn't removed this field. I'm fixing it.
Thank you.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
```diff
@@ -129,99 +126,107 @@ public void removeEventListener(EventListener listener) {
...
     /**
-     * Helper method that returns the current number of subscribed
-     * listeners.
+     * Helper method that returns the current number of subscribed listeners.
+     *
      * @return The current number listeners.
      */
     public int getListenerCount() {
-        return listeners.size();
+        int size = 0;
```
lucidd: Is there a reason you introduce this variable and not just return `listeners.size()` from inside the `synchronized` block?

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
```diff
@@ -3,113 +3,109 @@
...
-                        for (EventListener listener : listeners) {
-                            listener.notify(inboundEvent.readData(String.class));
-                        }
-                    }
-                }
+     * Connect the WebSocket to the server pointing to /ws/{token} to receive events.
+     */
+    private void initializeStream (ClientConfig config) throws SaltStackException {
+        try {
+            URI uri = new URI(config.get(ClientConfig.URL).toString()
```
lucidd: I think something like this should be more robust than string replacement.
```java
new URI("ws", uri.getSchemeSpecificPart(), uri.getFragment())
    .resolve("/ws/" + config.get(ClientConfig.TOKEN)))
```
In the long term i think it makes sense to have only hostname and port in the config instead of a whole URI. But that is out of scope for this PR. :smile:

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
```diff
@@ -3,113 +3,109 @@
...
      * Listeners that are notified of a new events.
      */
     private final List<EventListener> listeners = new ArrayList<>();
 
     /**
-     * The ClientConfig used to create the GET request for /events
+     * The {@link WebSocketContainer} object for a @ClientEndpoint implementation.
      */
-    private ClientConfig config;
+    private WebSocketContainer websocketContainer =
```
lucidd: This could be made `final`.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
```diff
@@ -129,99 +126,107 @@ public void removeEventListener(EventListener listener) {
...
+        return size;
+    }
+
+    /**
+     * Helper method to check if the WebSocket Session exists and is open.
+     *
+     * @return A flag indicating the {@link EventStream}
+     * WebSocket {@link Session} state.
+     */
+    public boolean isEventStreamClosed() {
```
lucidd:
```java
return this.session == null || !this.session.isOpen();
```
feels easier to read "the event stream is closed if session is null or session is not open."

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
```diff
@@ -129,99 +126,107 @@ public void removeEventListener(EventListener listener) {
...
+     * @param event The message received on this WebSocket.
      */
-    public boolean isEventProcessingStarted() {
-        return eventProcessingStarted;
+    @OnMessage
+    public void onMessage(String event) {
+        if (event != null && !event.equals("server received message")) {
+            synchronized (listeners) {
+                for (EventListener listener : listeners) {
+                    listener.notify(event.trim());
```
hustodemon: Hi!

Minor suggestion: we try to use some coolness from Java 8. I'd like to see it in the new code too. Could we do something like this
```java
listeners.stream().forEach(l -> l.notify(event.trim()));
```
here maybe?
(there's also similar thing on line 169)

lucidd: Is there a reason you `trim` the event string?

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
```diff
@@ -3,113 +3,109 @@
...
-                    }
-                }
+     * Connect the WebSocket to the server pointing to /ws/{token} to receive events.
+     */
+    private void initializeStream (ClientConfig config) throws SaltStackException {
+        try {
+            URI uri = new URI(config.get(ClientConfig.URL).toString()
+                    .replace("http", "ws") + "/ws/"
+                    + config.get(ClientConfig.TOKEN));
+            synchronized (websocketContainer) {
```
lucidd: I don't think this needs to be synchronized since its only called from the constructor and not accessible from outside.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
renner: Agreed, I was about to complain about the missing brackets (`{` `}`) in the `if` statement, but this is even better. But please put those brackets for all `if` statements (I have to add a checkstyle rule to enforce that).

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
ncounter: Hi. I'm realizing that you solution is safer not for styling code but for the variable out of synchronized block. I'm fixing it. Thank you.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
ncounter: Yes I was just replacing, but more robust code and no cons it means better solution to me. I like this.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
ncounter: I agree.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
ncounter: I'm fixing the code with this.
Sure, I will take care about brackets; I'm checking if any other statement needs them too.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
ncounter: @hustodemon Thank you for your suggestion, i'm changing the code according to Java 8.
@lucidd Yes, I <code>trim</code> the event string because it ends with <code>\n\n</code>

PR #89, opened 2015-08-07
Trivial changes

AdamMajer:
These get rid of compiler warnings.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
+++ b/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
@@ -61,7 +61,7 @@ public class ClientConfig {
 
     }
 
-    private final Map<Key, Object> store = new HashMap<>();
+    private final Map<Key<?>, Object> store = new HashMap<>();
 
     /**
      *  Sets the config for a key to the specified value.
--- a/src/main/java/com/suse/saltstack/netapi/event/EventStream.java
+++ b/src/main/java/com/suse/saltstack/netapi/event/EventStream.java
@@ -98,7 +98,7 @@ public class EventStream implements AutoCloseable {
                     uri.getSchemeSpecificPart(), uri.getFragment())
                     .resolve("/ws/" + config.get(ClientConfig.TOKEN));
             websocketContainer.setDefaultMaxSessionIdleTimeout(
-                    (long) config.get(ClientConfig.SOCKET_TIMEOUT));
+                    config.get(ClientConfig.SOCKET_TIMEOUT));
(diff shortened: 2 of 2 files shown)


renner:
Hello and thanks for your pull request! I am fine with using a wildcard type to fix that warning, saw that as well. About your other commit (49d3ab6) I was looking into it yesterday as well, you are using eclipse right? The point is that the code should actually compile fine since integers can safely be converted to longs automatically by the compiler. So the cast should not actually be necessary and the error that you see rather seems to be a [bug in eclipse] (https://bugs.eclipse.org/bugs/show_bug.cgi?id=440019). See also this [stackoverflow discussion] (http://stackoverflow.com/questions/27105299/java-8-converting-an-integer-to-a-long-compilation-issue).

However, even updating to 4.5 did not fix the issue for me, instead I get even more errors than before! Somehow it seems they didn't fix the bug completely or I might be missing something. Could you maybe try to use 4.5 and report what you see or is your eclipse on 4.5 already?

AdamMajer:
I was using version 4.4. Version 4.5, just installed, also has the same issue. Furthermore eclipse code cleanup then proceeds to remove this cast on save (happens with both versions). There is a [new bug about this](https://bugs.eclipse.org/bugs/show_bug.cgi?id=464496) that seems to be specific to generics.

I'm aware that widening convention shouldn't need explicit casts. The main reason I added this cast was actually to be consistent with code few lines down. On line 101 there is an explicit cast to `long` and it got rid off eclipse noise..

I guess I will remove the cast on 101 and reupload.

And yes, 4.5 gives new errors! It's not boxing up boolean -> Boolean automatically for shorthand lambda notation. It seems that,

    return executor.submit(() -> syndEvent(eventTag, eventData));

causes a problem, but

    return executor.submit(() -> { return syndEvent(eventTag, eventData); });

works around eclipse compiler's bug...

renner:
I confirm that. In general I think we should not work around eclipse bugs in our code, rather I'd like to see it fixed in eclipse itself soon. I hope that this will actually happen, otherwise we might have this patch in our stash for a while. Feel free to push a new version of your patch though removing the other cast as well (line 101) :+1:  Thank you!

AdamMajer:
Indeed. Working around ecj's bugs is probably not warranted if it's not one of the supported compilers.

Patches updated.

I've found another trivial thing to add to the trivial changes - missing `@Override` on implemented `AutoCloseable.close()`

renner:
We could work around those issues, but actually I guess that whatever compiles with maven and javac should also compile in eclipse, right? Merging it, thank you @AdamMajer!

Issue #83, opened 2015-07-04
Support job status for unfinished jobs


mchugh19:
When kicking off a job, the saltapi response includes a minion array listing minion names and the job results. If the job is long running,  you can iterate through the number of minions that are targeted, and spin waiting for all to return. 

Please support querying this minion listing.

mchugh19:
To clarify, when you send a long running job to salt, like in this example using sleep:
```json
[{
    "arg": ["sleep 14; echo testing"],
    "client": "local",
    "expr_form": "glob",
    "fun": "cmd.run",
    "tgt": "dynocache2*"
}]
```
you get back a JID. When you lookup the job using that JID, you get back:

```json
{
    "info": [{
        "Function": "cmd.run",
        "jid": "20150704153935653922",
        "Target": "dynocache2*",
        "Target-type": "glob",
        "User": "jenkins",
        "StartTime": "2015, Jul 04 15:39:35.653922",
        "Arguments": ["sleep 14; echo testing"],
        "Minions": ["dynocache2v2.cws.dpdcn.unt.edu", "dynocache2v1.cws.dpdcn.unt.edu"],
        "Result": {}
    }],
    "return": [{}]
}
```

Note that the result, and return keys are empty, but the minion key is populated with the hosts we are waiting on. Thus, even though the job has not yet finished, we are able to see that we are waiting for two hosts in this example.

Once the job has completed, looking up that JID now returns:
```json
{
    "info": [{
        "Function": "cmd.run",
        "jid": "20150704153935653922",
        "Target": "dynocache2*",
        "Target-type": "glob",
        "User": "jenkins",
        "StartTime": "2015, Jul 04 15:39:35.653922",
        "Arguments": ["sleep 14; echo testing"],
        "Minions": ["dynocache2v2.cws.dpdcn.unt.edu", "dynocache2v1.cws.dpdcn.unt.edu"],
        "Result": {
            "dynocache2v2.cws.dpdcn.unt.edu": {
                "return": "testing"
            },
            "dynocache2v1.cws.dpdcn.unt.edu": {
                "return": "testing"
            }
        }
    }],
    "return": [{
        "dynocache2v2.cws.dpdcn.unt.edu": "testing",
        "dynocache2v1.cws.dpdcn.unt.edu": "testing"
    }]
}
```

Notice here we have both the return key and the result key populated with our expected results so we can tell that both of our hosts have come back. 

[...]

PR #90, opened 2015-08-07
{WIP} Add pending jobs to JID result set

AdamMajer:
Fixes: #83 

This is work-in-progress. I would appreciate some feedback since this issue can be fixed in many ways.

The approach in this patch adds JobsState.PENDING enum that is artificially added to results set for minions that have yet to return. This may be a little blunt. API is not changed, but user now has to check if job is pending before processing minion's output.

An alternative is to parse `info` and disregard `return`. `info["Result"]` seems to contain the same data as in `return`, though in slightly different order. Returning Info instead of a result map would definitely result in API changes, but it would allow for much more flexible interface for job results.

What do you guys think?


AdamMajer:
I'll move discussion to issue area. Seems github references it all over the place!

PR #92, opened 2015-08-11
Correct order of some unit tests assertEquals() parameters

AdamMajer:
The correct order is assertEquals(expected, actual) and some unit
tests were using reverse order. Using incorrect order makes any test
failure messages somewhat confusing.


lucidd:
Thanks for fixing this. The patch looks good. :+1: 
(i wonder if i'm the only one that finds the order of assertEquals confusing :smile: )

AdamMajer:
It's probably confusing to anyone that used other test packages. For example, in CUnit, `CU_ASSERT_EQUAL(actual, expected)`. or in QTest, [QCOMPARE( actual, expected)](http://doc.qt.io/qt-5/qtest.html#QCOMPARE)

renner:
Thanks for these fixes, @AdamMajer! Merged.

PR #93, opened 2015-08-14
Improve job result handling

AdamMajer:
These two patches attempt to address the "pending job" enhancement (#83)

The first patch simply adds capability to parse the `info` structure. This structure contains duplicate minion result set. I've included a convenience function to return result set in the same format as the old `return` structure.

The second patch replaces old job result and returns ResultInfoSet instead of a simple map. The simple map is still available from within the ResultInfo class.

### DIFF
--- a/src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
+++ b/src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
@@ -15,6 +15,8 @@ import com.suse.saltstack.netapi.datatypes.Keys;
 import com.suse.saltstack.netapi.datatypes.ScheduledJob;
 import com.suse.saltstack.netapi.datatypes.Token;
 import com.suse.saltstack.netapi.utils.ClientUtils;
+import com.suse.saltstack.netapi.results.ResultInfo;
+import com.suse.saltstack.netapi.results.ResultInfoSet;
 
 import static com.suse.saltstack.netapi.config.ClientConfig.SOCKET_TIMEOUT;
 import static com.suse.saltstack.netapi.AuthModule.PAM;
@@ -23,6 +25,7 @@ import static com.suse.saltstack.netapi.AuthModule.AUTO;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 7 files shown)


renner:
Hey Adam, thanks a lot for your efforts! Your patch seems well done. I wonder how this merges with the changes @lucidd is currently working on in [his branch] (https://github.com/SUSE/saltstack-netapi-client-java/compare/redesign). @lucidd, could we actually merge this PR and rebase your branch on top? At least `getjobResults()` is not affected by the changes in your branch and it would be nice to return something different here than `Map<String, Object>`.

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
```diff
@@ -706,6 +714,38 @@ public void testJobsAsync() throws Exception {
...
+        assertNotNull(results);
+        assertEquals(0, results.getResults().size());
+        assertTrue(!results.getResult("mira").isPresent());
+        assertEquals("cmd.run", results.getFunction());
+        assertEquals("*", results.getTarget());
+        assertEquals("adamm", results.getUser());
+        assertEquals(pending_minions, results.getMinions());
+        assertEquals(pending_minions, results.getPendingMinions());
+        assertEquals(Date.from(Instant.parse("2015-08-06T16:55:13.400200Z")),
+                results.getStartTime());
```
renner: This assertion actually fails for me, probably due to my local timezone setting. Could we format the date using `DATE_FORMAT` for this comparison as we do it in `testJobs()`?

```java
assertEquals("2015-08-06 16:55:13", DATE_FORMAT.format(results.getStartTime()));
```

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
```diff
@@ -706,6 +714,38 @@ public void testJobsAsync() throws Exception {
...
 
     @Test
+    public void testJobsPending() throws Exception {
+        stubFor(any(urlMatching(".*"))
+                .willReturn(aResponse()
+                .withStatus(HttpURLConnection.HTTP_OK)
+                .withHeader("Content-Type", "application/json")
+                .withBody(JSON_JOBS_RESPONSE_PENDING)));
+
+        ResultInfoSet result_set = client.getJobResult("some-job-id");
```
renner: Let's use mixed case with a lowercase first letter for variables here (`resultSet`) and below (`pendingMinions`).

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
AdamMajer: Fixed.

And yes, it is related to TZ values and I think it shows a little bit of weakness in parsing code. Date is always returned relative to local TZ - TZ should probably be configurable via ClientConfig if TZ data is not available for the server.

src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
AdamMajer: This should be all done now. I'm too used to using lower_case for variables and camelCase for functions and CamelCase for other types :smile:

renner:
We merged it, thanks a lot. I just pushed two fixes on top: a83dd2672ab582e621d8a0ea3ecb44a788ccc2a7 and f6863a1edc5406899512af46e2d93022966b8595.

PR #94, opened 2015-08-17
Support for calling salt function in wheel, runner and execution modules.

lucidd:
This adds basic support to call salt function in wheel, runner and execution modules in a seamless way.


renner:
Ok, I'm done with cleaning up this patch for now, can be merged from my perspective.

isbm:
@renner is usually honest at describing complete rewrites... :laughing:

renner:
Merged after rebasing on top of a more recent master.

PR #95, opened 2015-08-17
Basic parsing of events

renner:
This patch brings in basic parsing of events into objects. So far we are only separating an event's `tag` from its `data` though. This is already very helpful in order to filter the event stream for interesting events. Apart from that there is two small commits affecting the event stream tests. Please have a look, @lucidd, thank you!

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/datatypes/Event.java
+++ b/src/main/java/com/suse/saltstack/netapi/datatypes/Event.java
@@ -1,0 +1,28 @@
+package com.suse.saltstack.netapi.datatypes;
+
+import java.util.HashMap;
+
+/**
+ * Parse events into objects.
+ */
+public class Event {
+
+    private String tag;
+    private HashMap<String, Object> data;
+
+    /**
+     * Return this event's tag.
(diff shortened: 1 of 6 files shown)


src/main/java/com/suse/saltstack/netapi/datatypes/Event.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.util.HashMap;
+
+/**
+ * Parse events into objects.
+ */
+public class Event {
+
+    private String tag;
+    private HashMap<String, Object> data;
```
lucidd: i would just use `Map` here no reason to be more specific.

lucidd:
Besides the lines i commented on LGTM :+1:

src/main/java/com/suse/saltstack/netapi/datatypes/Event.java
renner: Thanks, fixed with 9aec7a51ff7c392d671e6fda4ea41d4a2c7a7ee7.

PR #101, opened 2015-09-02
Add timezone support for parsing StartTime

AdamMajer:
This attempts to add TimeZone configuration support for master's StartTime parsing. By default, if TimeZone is unset and left as system default, there should be no performance impact.

To facilitate per-server timezone settings, parser needs to be aware of configuration. It appears that annotations are unworkable in this scenario because Gson uses reflection to to create instance for annotated types and these do not have access to JsonParser instance. No JsonParser instance, no configuration, no timezones. I've added a simple StartTime wrapper class to remove need for annotations. External API is not affected.

There appears to be no way around rebuilding a parser though. But this only affects users that have different timezone from server (master), which I'd expect to be mostly desktop applications.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/datatypes/StartTime.java
+++ b/src/main/java/com/suse/saltstack/netapi/datatypes/StartTime.java
@@ -1,0 +1,80 @@
+package com.suse.saltstack.netapi.datatypes;
+
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Locale;
+import java.util.TimeZone;
+
+import com.suse.saltstack.netapi.exception.ParsingException;
+
+/**
+ * StartDate is a convenience wrapper allowing for parsing of StartDate in the
+ * timezone appropriate to a given master.
+ */
+
+public class StartTime {
+    private final String dateString;
+    private transient final Date defaultTzDate;
+
+    // StartTime example from API: "2015, Mar 04 19:28:29.724698"
+    private transient final SimpleDateFormat startJobDateFormat =
+            new SimpleDateFormat("yyyy, MMM dd HH:mm:ss.SSS", Locale.US);
+
+    public StartTime(String dateString) throws ParsingException {
+        try {
+            if (dateString != null)
+                this.defaultTzDate = startJobDateFormat.parse(dateString);
+            else
+                this.defaultTzDate = null;
+        } catch (ParseException e) {
+            throw new ParsingException(e);
+        }
+        this.dateString = dateString;
+    }
+
+    /**
+     * Returns {@link Date} representation of StartTime as parsed at a given timezone.
+     * Master does not return a timezone associated with StartTime timestamp string,
+     * therefore an explicit timezone needs to be provided for correct parsing.
+     *
+     * @param tz TimeZone associated with master.
+     * @return {@link Date} representation of StartTime at provided timezone
+     */
+    public synchronized Date getDate(TimeZone tz) {
+        if (dateString == null)
+            return null;
+
+        startJobDateFormat.setTimeZone(tz);
+        try {
+            return startJobDateFormat.parse(dateString);
+        } catch (ParseException e) {
+            // This exception should not be possible. Even when dealing with TZ DST
+            // transitions, all dates exist. Time that does not exist because of
+            // forward DST transition, get automatically forwarded during parsing.
+            throw new RuntimeException("Internal problem with Java and timezones " +
+                    dateString + " @TZ: " + tz.toString());
+        }
+    }
+
+    /**
+     * Returns {@link Date} representation of StartTime as parsed using default timezone.
+     *
+     * <p>NOTE: If master is using a different timezone than the default timezone of the
+     * user of this API, then the returned Date will be incorrect.
+     *
+     * @return {@link Date} representation of StartTime using default timezone.
+     */
+    public Date getDate() {
+        return defaultTzDate;
+    }
+
+    /**
+     * Returns a string representation of StartTime. This is the same string that is
+     * passed to the constructor.
+     */
+    @Override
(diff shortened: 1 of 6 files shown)


AdamMajer:
I've added additional synchronization for SimpleDateTime. Alternatively, we could use Apache's library that works better with threads,

http://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/time/FastDateFormat.html

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -77,9 +87,54 @@ public JsonParser(TypeToken<T> type) {
...
+        this.gson = new GsonBuilder()
+                .registerTypeAdapter(Date.class, new DateAdapter().nullSafe())
+                .registerTypeAdapter(StartTime.class,
+                        new JobStartTimeJsonAdapter().nullSafe())
+                .registerTypeAdapter(Stats.class, new StatsDeserializer())
+                .registerTypeAdapter(Arguments.class, new ArgumentsDeserializer())
+                .create();
+
+        synchronized (START_TIME_FORMAT) {
+            this.startJobDateFormat = (SimpleDateFormat) START_TIME_FORMAT.clone();
```
lucidd: Java `clone` should generally be avoided unless you know how to handle the demons this black magic might summon into this world.
Just create a helper method that takes all the relevant parts and puts them together into a new `SimpleDateFormat`.

AdamMajer: OK. I've changed this now. This entire copy constructor is actually useless and it arose from me not initially understanding the flow of the program (how SaltStackClient uses JsonParser)

The simplest approach is probably the best - 1 instance of format per JsonParser instance.

AdamMajer:
I've added caching when duplicating parsers for different TimeZones. Various parsers created as part of the new callSync/callAsync functions now also get correct TimeZone at creation time. JsonParser constructor was changed to require TimeZone to facilitate this.

Static Parsers, they are duplicated only on TimeZone change and then cached for duration. Parsers created dynamically via the call() interface, are not cached.

AdamMajer:
Rebased so it should merge again.

Any comments on this pull request? (is it good, bad?)

My main motivation is if master is in different TZ from client, startDate is parsed wrong. Other dates use unix-like timestamp, which get parsed correctly into the timezone, but startDate does not. So if an application starts to compare the two, there will be timezone issues.

lucidd:
@AdamMajer sorry for taking so long to answer. Your pull request looks good. The only reason why we haven't merged it yet is that we are not sure if it is worth adding all this complexity and changing the parser just to parse this one date properly. Another approach that we discussed is not parsing it at all but modifying the getter to parse it by providing it the right timezone as an argument but that would put more work on the user of the API.

AdamMajer:
Indeed. That approach may indeed be better. In that case, startTime becomes just a String internally.

User of the API already has to either know the timezone of master(s), or use the timezone module if master runs minions. This means there wouldn't really be additional workload on client to request TZ in startTime getter.

I'll rework this then to move this TZ handling to getter.

AdamMajer:
This is now much simpler. The StartTime class remains, and it's now more useful considering it is parsing in the getter.

The constructor is also doing a parse with default TZ purely to catch parse exceptions and forward them to the JsonParser so it works like before with respect to exceptions. The exception in the getter should never actually be called hence I've changed it to an unchecked exception.

Synchronization problems in original StartTime parser are now also fixed as a side-effect of these changes.

src/main/java/com/suse/saltstack/netapi/datatypes/StartTime.java
```diff
@@ -0,0 +1,65 @@
...
+        this.dateString = dateString;
+    }
+
+    /**
+     * Returns @{link Date} representation of StartTime as parsed at a given timezone.
+     * Master does not return a timezone associated with StartTime timestamp string,
+     * therefore an explicit timezone needs to be provided for correct parsing.
+     *
+     * @param tz TimeZone associated with master.
+     * @return @{link Date} representation of StartTime at provided timezone
```
renner: The `@` needs to go inside the `{}` here (and above), otherwise javadoc will fail thinking that this is the next tag, see [travis] (https://travis-ci.org/SUSE/saltstack-netapi-client-java/builds/79381432#L1574).

renner:
I like this version of the patch much better! :+1: I was wondering though: would it make sense to make the timezone parameter in the getter optional? Could we keep the old getter for instance (no parameter) and it would return the job start time either in the timezone of the caller or the master? This would make it a bit easier for users who have their master and minions in the same timezone since they wouldn't need to hard-code their timezone into the calls.

AdamMajer:
I was under the impression that StartTime's timezone was determined by the master and it was irrelevant where the minions where. From the data structure that would make sense - one StartTime and many Minions. Did you mean master in same timezone as user of the library?

But yes, TimeZone.getDefault() case getting would make sense too, considering that case is parsed in the constructor!

renner:
Yes, sorry, I meant salt-master in the same timezone as the user of the library.

AdamMajer:
Updated now to use default TimeZone, if none is supplied.

But from bug-prone point of view, I would prefer if this option didn't exist. Start date is pretty much the only (?) date representation that can be completely wrong without a time zone. Users may not know about this quirkiness and assume that Date is a Date. By requiring a time zone, at very least they would pause to think "why is TZ required?". And if specific time zone is not required, then TimeZone.getDefault() is not difficult.

renner:
I merged it, thank you! Point taken, but I tend to prefer to keep the no arguments getter around. Further I fixed up some minor issues in the `StartTime` class with 2067c23ff7594291fbd4977f632e3dcf5ff7f286. Thanks again for your contributions and sorry that it took us quite a while this time to get it merged!

PR #102, opened 2015-09-02
Add support for all key states in list_all and finger

lucidd:


lucidd:
@AdamMajer good catch that should be static indeed thanks mate :smile:

renner:
Merged.

PR #105, opened 2015-09-07
Add support for salt.modules.status

renner:
This patch adds support for using the functions from the [`salt.modules.status` module] (https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.status.html) with the `callSync`/`callAsync` interfaces. All functions are represented in the new class, the only one I left out is [`master()`] (https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.status.html#salt.modules.status.master), which I was not sure if it makes sense to be called via the API.

I tested all of the return types manually, but they are still very generic for many functions. This can definitely be improved in terms of parsing the data into more specialized objects.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/calls/modules/Status.java
+++ b/src/main/java/com/suse/saltstack/netapi/calls/modules/Status.java
@@ -1,0 +1,108 @@
+package com.suse.saltstack.netapi.calls.modules;
+
+import com.suse.saltstack.netapi.calls.LocalCall;
+
+import com.google.gson.reflect.TypeToken;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * salt.modules.status
+ */
+public class Status {
+
+    private Status() { }
+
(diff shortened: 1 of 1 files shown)


src/main/java/com/suse/saltstack/netapi/calls/modules/Status.java
```diff
@@ -0,0 +1,110 @@
...
+                new TypeToken<String>(){});
+    }
+
+    public static LocalCall<String> version() {
+        return new LocalCall<>("status.version", Optional.empty(), Optional.empty(),
+                new TypeToken<String>(){});
+    }
+
+    public static LocalCall<Map<String, Long>> vmstats() {
+        LinkedHashMap<String, Object> args = new LinkedHashMap<>();
```
lucidd: copy paste mistake :smile:

lucidd:
Other then that looks good to me :+1:

src/main/java/com/suse/saltstack/netapi/calls/modules/Status.java
renner: Thanks! Fixed it.

renner:
Merged.

PR #106, opened 2015-09-07
SimpleDateFormat from static variable to local

mbologna:
Moved SimpleDateFormat from static variable to local variable inside the
method they are used, because DateFormats are not thread-safe, meaning that
they maintain an internal representation of state. Using them in a static
context can yield some pretty strange bugs if multiple threads access the
same instance simultaneously.

Reference:
https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html#synchronization

### DIFF
--- a/src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
+++ b/src/test/java/com/suse/saltstack/netapi/client/SaltStackClientTest.java
@@ -608,7 +608,7 @@ public class SaltStackClientTest {
     @Test
     public void testJobs() throws Exception {
     	final SimpleDateFormat DATE_FORMAT =
-            new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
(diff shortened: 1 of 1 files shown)


AdamMajer:
You are absolutely correct that SimpleDateFormat is not thread-safe. That is one of the reasons for FastDateFormat in Apache's commons library.

But does jUnit now run tests in parallel by default? I thought they were still sequential.

mbologna:
They are run in sequential by default.
From JUnit 4.7 (and surefire maven plugin) they could be run in parallel mode.

renner:
Hey, thanks for this patch! I would like to merge it, but you could you please fix the formatting again and use four spaces instead of tabs? Your patch introduces four tabs (right before the `final` and `new` keywords).

renner:
Thanks, merged.

mbologna:
Thanks!

PR #107, opened 2015-09-08
Add dependency on slf4j-jdk14 logger

AdamMajer:
wiremock is pulling in slf4j as its dependency in unit tests, but
because it does not pull in any logging implementation, wiremock's
logs end up in /dev/null.

...
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
...

Adding a JDK logging backend for slf4j gives us wiremock logs.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,6 +129,12 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-jdk14</artifactId>
+      <version>1.7.6</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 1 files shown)


renner:
Ok, we merged it after updating the dependency to the latest version, see 42522a076e7fbb05ea278176b759ea8aa453f090. We already fixed a first bug that was discovered via the new logging (acaf0c7649d0718a6c49c491c7b365425ae2bc30), thanks a lot for this contribution!

AdamMajer:
You are welcome!

PR #109, opened 2015-09-08
add pkg.info to Pkg module

lucidd:
This i work in progress and not ready to be merged yet since attribute names might change upstream.


src/main/java/com/suse/saltstack/netapi/calls/modules/Pkg.java
```diff
@@ -43,6 +45,119 @@ public String getSummary() {
         }
     }
 
+    /**
+     * Information about a package as returned by pkg.info_installed and
+     * pkg.info_available
+     */
+    public static class Info {
```
renner: Are you sure this should be named just `Info`? Wouldn't something like `PackageInfo` make it more clear what this is?

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -268,4 +315,28 @@ public StartTime read(JsonReader jsonReader) throws IOException {
             return new StartTime(subStr);
         }
     }
+
+
+    /**
+     *
+     */
```
renner: Please either remove this empty comment or add some documentation to this class.

renner:
Looks good to me! :+1: Feel free to merge as soon as you are ready.

src/main/java/com/suse/saltstack/netapi/calls/modules/Pkg.java
renner: Nevermind, I see now that it should be used as `Pkg.Info`, makes sense.

lucidd: The idea here is similar to what we have with `Key` that the module name is part of the path you specify when referring to the type.

```java
import com.suse.saltstack.netapi.calls.modules.Pkg;

Pkg.Info a = ...
```

PR #110, opened 2015-09-08
make argument types easier to work with

lucidd:


renner:
Merged it, thanks for this patch!

PR #111, opened 2015-09-09
On event stream closure, pass the CloseReason on to the listeners

renner:
This patch makes it easier to debug the websocket based event stream by passing on the [CloseReason] (http://docs.oracle.com/javaee/7/api/javax/websocket/CloseReason.html) to all event listeners whenever the event stream has closed.


renner:
Merged after positive feedback from @lucidd.

PR #113, opened 2015-09-16
add SaltVersion to represent and parse salt version strings

lucidd:
add SaltVersion to represent and parse salt version strings :stuck_out_tongue:

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/datatypes/SaltVersion.java
+++ b/src/main/java/com/suse/saltstack/netapi/datatypes/SaltVersion.java
@@ -1,0 +1,152 @@
+package com.suse.saltstack.netapi.datatypes;
+
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Represents a salt version
+ */
+public class SaltVersion implements Comparable<SaltVersion> {
+
+    private static Pattern SALT_VERSION_REGEX =
+            Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)([Rr][Cc](\\d+))?$");
+
+    private final int year;
+    private final int month;
+    private final int bugfix;
+    private final Optional<Integer> releaseCandidate;
+
+    /**
+     * Parses a salt version string
+     * @param versionString the salt version string
+     * @return SaltVersion if the versionString is valid or empty Optional if not.
+     */
+    public static Optional<SaltVersion> parse(String versionString) {
+        Matcher matcher = SALT_VERSION_REGEX.matcher(versionString);
+        if (matcher.matches()) {
+            int year = Integer.parseInt(matcher.group(1));
+            int month = Integer.parseInt(matcher.group(2));
+            int bugfix = Integer.parseInt(matcher.group(3));
+            Optional<Integer> rc = Optional.ofNullable(matcher.group(5))
+                    .map(Integer::parseInt);
+            return Optional.of(new SaltVersion(year, month, bugfix, rc));
+        } else {
+            return Optional.empty();
+        }
+    }
+
+    /**
+     * Creates a SaltVersion
+     * @param year Year of the release
+     * @param month Month of the release
+     * @param bugfix Bugfix number incremented withing a feature release
+     * @param releaseCandidate Optional release candidate tag
+     */
+    public SaltVersion(int year, int month, int bugfix,
+            Optional<Integer> releaseCandidate) {
+        this.year = year;
+        this.month = month;
+        this.bugfix = bugfix;
+        this.releaseCandidate = releaseCandidate;
+    }
+
+    /**
+     * Creates a SaltVersion
+     * @param year Year of the release
+     * @param month Month of the release
+     * @param bugfix Bugfix number incremented withing a feature release
+     * @param releaseCandidate release candidate tag
+     */
+    public SaltVersion(int year, int month, int bugfix, int releaseCandidate) {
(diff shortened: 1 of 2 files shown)


lucidd:
linking to the discussion here https://github.com/SUSE/saltstack-netapi-client-java/commit/b67a55f since its not visible here anymore. @AdamMajer please always try to comment on the Files changed tab and not on the commit itself otherwise those comments are not linked to the PR anymore if the commit gets rewritten.

lucidd:
Also its not a problem if you comment a lot  every message / feedback is appreciated :smile:

renner:
@lucidd, could you please always provide a description, I mean for **every** PR? If the commit message is really enough information please just copy it over to the description, still better than "No description provided" :wink:

src/main/java/com/suse/saltstack/netapi/datatypes/SaltVersion.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Represents a salt version
+ */
+public class SaltVersion implements Comparable<SaltVersion> {
+
+    private static Pattern SALT_VERSION_REGEX =
+            Pattern.compile("^(\\d+)\\.(\\d+)\\.(\\d+)(rc(\\d+))?$");
```
isbm: @lucidd  `....(rc(...` vs `....([rcRC](...` ? Otherwise LGTM.

lucidd: how about `[Rr][Cc]` so you can mix it however you like :smile:

AdamMajer:
Thank you @lucidd for the github tip. Github does certainly functions differently from gerrit with respect to change tracking.

Cursory look doesn't bring up any errors for me.

With respect to earlier comment about exception vs. Optional for parse errors, I think exceptions allow for easier detection of any future incompatibilities with Salt. If version string gets a different format (for example, like Linux went from 3 version numbers to 2 for releases (+1 for patches)), then failing loudly at point of failure makes it easier to track down instead of "version went blank".

lucidd:
@AdamMajer it will still fail very loudly if you want it to just call `.get` on the option and you will get the exception. `Option` here does not mean version is blank but rather "you gave ma an invalid version string so i don't have a version object for you".

AdamMajer:
I think I'll leave it for @renner to decide if parse errors are handled satisfactorily here. As an example, `SimpleDateFormat.parse()` also returns null (Optional.empty() equivalent) on invalid parse so your code is not unique in that.

I see no coding errors, so :+1:

src/main/java/com/suse/saltstack/netapi/datatypes/SaltVersion.java
```diff
@@ -0,0 +1,145 @@
...
+     * @param versionString the salt version string
+     * @return SaltVersion if the versionString is valid or empty Optional if not.
+     */
+    public static Optional<SaltVersion> parse(String versionString) {
+        Matcher matcher = SALT_VERSION_REGEX.matcher(versionString);
+        if (matcher.matches()) {
+            int year =  Integer.parseInt(matcher.group(1));
+            int month =  Integer.parseInt(matcher.group(2));
+            int bugfix =  Integer.parseInt(matcher.group(3));
+            Optional<Integer> rc =  Optional.ofNullable(matcher.group(5))
```
isbm: Am I just see here extra spaces from 28 to 31st line?

src/main/java/com/suse/saltstack/netapi/datatypes/SaltVersion.java
```diff
@@ -0,0 +1,145 @@
...
+    private final int month;
+    private final int bugfix;
+    private final Optional<Integer> releaseCandidate;
+
+    /**
+     * Parses a salt version string
+     * @param versionString the salt version string
+     * @return SaltVersion if the versionString is valid or empty Optional if not.
+     */
+    public static Optional<SaltVersion> parse(String versionString) {
```
AdamMajer: I'm sorry, but now I remember what bothered me about this the most. The Optional<> here is completely unnecessary. It doesn't add anything except complicates the API. Returns null on invalid and return SaltVersion on valid. Drop Optional<> wrapper. If someone wants to wrap this in Optional for some reason, they can do this via `Optional.ofNullable(SaltVersion.parse(str));` in their code.

lucidd: The whole idea of `Optional` is to get rid of null and replace it with a safer alternative. What `Optional` does is it encodes the concept of something being present or not in the type itself. Similar to a `List` that encodes the possibility of multiple values being there. Think of it as List is a way to say there are 0 or more values while `Optional` is 0 or 1 value. The advantage of encoding things in the type is that the developer immediately sees whats going on without reading the docs. In contrast to null you have to actively deal with `Optional` by calling `map` , `ifPresent`, `orElse` or `getOrElse` (`get` excluded because its not safe and should not be used) otherwise you will not get access to its inner value if there is one. With those helper methods on `Optional` you can write your logic as if the value is there and `Optional` will do the "null checking" for you.

AdamMajer: I think you misunderstand. I know what `Optional` does and I've written some small functional things (in Erlang, for example). My comment is that `Optional` wrapper belongs in the caller-code, not the called-code. The null checking, be that via `Optional` or otherwise, is done in caller-code anyway.

Anyway, this is all done and merged now. I just want to clarify any misunderstandings about my statements.

PR #114, opened 2015-09-17
Buffer partial websocket messages

renner:
This patch adds support for buffering partial websocket messages in `EventStream` by implementing [`MessageHandler.Partial`] (http://docs.oracle.com/javaee/7/api/javax/websocket/MessageHandler.Partial.html). A configurable limit on the allowed message length is introduced that can be disabled by setting a value less than or equal to 0. Partial messages will be buffered and listeners get notified only when the last message has been received. The stream will be closed with a suitable [`CloseReason`] (http://docs.oracle.com/javaee/7/api/javax/websocket/CloseReason.html) if the buffer grows beyond the maximum length.

This is what you get without this patch using Tomcat as client and receiving a "big" event:
```
Event stream closed: The decoded text message was too big for the output buffer and the endpoint does not support partial messages [TOO_BIG]
```

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
+++ b/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
@@ -37,7 +37,7 @@ public class ClientConfig {
     /**
      * Maximum websocket message length in characters. The default value corresponds to
      * 10 MB, assuming that one character takes up 2 bytes. This limit can be disabled
-     * by setting a value <= 0.
+     * by setting a value less than or equal to 0.
      */
     public static final Key<Integer> WEBSOCKET_MAX_MESSAGE_LENGTH = new Key<>(5242880);
 
--- a/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
+++ b/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
@@ -39,7 +39,7 @@ public class ClientConfig {
      * 10 MB, assuming that one character takes up 2 bytes. This limit can be disabled
      * by setting a value less than or equal to 0.
      */
-    public static final Key<Integer> WEBSOCKET_MAX_MESSAGE_LENGTH = new Key<>(5242880);
+    public static final Key<Integer> WEBSOCKET_MAX_MESSAGE_LENGTH = new Key<>(0x500000);
 
     /**
      * A key to use with {@link ClientConfig}.
--- a/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
+++ b/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
@@ -35,6 +35,13 @@ public class ClientConfig {
     public static final Key<String> PROXY_PASSWORD = new Key<>();
 
     /**
+     * Maximum websocket message length in characters. The default value corresponds to
+     * 10 MB, assuming that one character takes up 2 bytes. This limit can be disabled
+     * by setting a value <= 0.
+     */
+    public static final Key<Integer> WEBSOCKET_MAX_MESSAGE_LENGTH = new Key<>(5242880);
+
+    /**
      * A key to use with {@link ClientConfig}.
      * @param <T> The type of the value associated with this key.
      */
--- a/src/main/java/com/suse/saltstack/netapi/event/EventStream.java
+++ b/src/main/java/com/suse/saltstack/netapi/event/EventStream.java
@@ -40,6 +40,16 @@ public class EventStream implements AutoCloseable {
     private final List<EventListener> listeners = new ArrayList<>();
 
     /**
+     * Default message buffer size in characters.
+     */
+    private final int defaultBufferSize = 1024;
+
+    /**
+     * Buffer for partial messages.
+     */
+    private final StringBuilder partialMessageBuffer = new StringBuilder(defaultBufferSize);
+
+    /**
      * The {@link WebSocketContainer} object for a @ClientEndpoint implementation.
      */
     private final WebSocketContainer websocketContainer =
@@ -183,18 +193,37 @@ public class EventStream implements AutoCloseable {
     }
 
     /**
-     * Notify listeners on each event received on the WebSocket.
+     * Notify listeners on each event received on the websocket and buffer partial messages.
      *
-     * @param message The message received on this WebSocket
+     * @param partialMessage partial message received on this websocket
+     * @param last indicate the last part of a message
      */
     @OnMessage
-    public void onMessage(String message) {
-        if (message != null && !message.equals("server received message")) {
-            // Salt API adds a "data: " prefix that we need to ignore
-            Event event = JsonParser.EVENTS.parse(message.substring(6));
-            synchronized (listeners) {
-                listeners.stream().forEach(l -> l.notify(event));
+    public void onMessage(String partialMessage, boolean last) {
+        if (last) {
+            String message;
+            if (partialMessageBuffer.length() == 0) {
+                message = partialMessage;
+            } else {
+                partialMessageBuffer.append(partialMessage);
+                message = partialMessageBuffer.toString();
             }
+
+            // Notify all registered listeners
+            if (!message.equals("server received message")) {
+                // Salt API adds a "data: " prefix that we need to ignore
+                Event event = JsonParser.EVENTS.parse(message.substring(6));
+                synchronized (listeners) {
+                    listeners.stream().forEach(l -> l.notify(event));
+                }
+            }
+
+            // Empty the buffer (and reset size to the defaultBufferSize)
+            partialMessageBuffer.setLength(0);
+            partialMessageBuffer.trimToSize();
+            partialMessageBuffer.ensureCapacity(defaultBufferSize);
+        } else {
+            partialMessageBuffer.append(partialMessage);
         }
     }
 
--- a/src/main/java/com/suse/saltstack/netapi/event/EventStream.java
+++ b/src/main/java/com/suse/saltstack/netapi/event/EventStream.java
@@ -216,7 +216,7 @@ public class EventStream implements AutoCloseable {
                 partialMessageBuffer.append(partialMessage);
                 message = partialMessageBuffer.toString();
             }
-            if (maxMessageLength >= 0 && message.length() > maxMessageLength) {
+            if (maxMessageLength > 0 && message.length() > maxMessageLength) {
                 throw new MessageTooBigException(maxMessageLength);
             }
 
@@ -235,7 +235,7 @@ public class EventStream implements AutoCloseable {
             partialMessageBuffer.setLength(0);
         } else {
             partialMessageBuffer.append(partialMessage);
-            if (maxMessageLength >= 0 && partialMessageBuffer.length() > maxMessageLength) {
+            if (maxMessageLength > 0 && partialMessageBuffer.length() > maxMessageLength) {
                 throw new MessageTooBigException(maxMessageLength);
             }
         }
--- a/src/main/java/com/suse/saltstack/netapi/event/EventStream.java
+++ b/src/main/java/com/suse/saltstack/netapi/event/EventStream.java
@@ -64,7 +64,7 @@ public class EventStream implements AutoCloseable {
     /**
      * The WebSocket {@link Session}.
      */
-    public Session session;
+    private Session session;
 
     /**
      * Constructor used to create this object.
--- a/src/main/java/com/suse/saltstack/netapi/event/EventStream.java
+++ b/src/main/java/com/suse/saltstack/netapi/event/EventStream.java
@@ -248,14 +248,9 @@ public class EventStream implements AutoCloseable {
      * @throws IOException in case of an error when closing the session
      */
     @OnError
-    public void onError(Throwable t) throws IOException {
-        CloseReason closeReason;
-        if (t instanceof MessageTooBigException) {
-            closeReason = new CloseReason(CloseCodes.TOO_BIG, t.getMessage());
-        } else {
-            closeReason = new CloseReason(CloseCodes.CLOSED_ABNORMALLY, t.getMessage());
-        }
-        close(closeReason);
+    public void onError(Throwable throwable) throws IOException {
+        close(new CloseReason(throwable instanceof MessageTooBigException ?
+                CloseCodes.TOO_BIG : CloseCodes.CLOSED_ABNORMALLY, throwable.getMessage()));
     }
 
     /**
--- a/src/main/java/com/suse/saltstack/netapi/event/EventStream.java
+++ b/src/main/java/com/suse/saltstack/netapi/event/EventStream.java
@@ -229,10 +229,10 @@ public class EventStream implements AutoCloseable {
                 }
             }
 
-            // Empty the buffer (and reset size to the defaultBufferSize)
-            partialMessageBuffer.setLength(0);
+            // Reset the size to the defaultBufferSize and empty the buffer
+            partialMessageBuffer.setLength(defaultBufferSize);
             partialMessageBuffer.trimToSize();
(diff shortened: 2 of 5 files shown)


src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
```diff
@@ -35,6 +35,13 @@
     public static final Key<String> PROXY_PASSWORD = new Key<>();
 
     /**
+     * Maximum websocket message length in characters. The default value corresponds to
+     * 10 MB, assuming that one character takes up 2 bytes. This limit can be disabled
+     * by setting a value <= 0.
+     */
+    public static final Key<Integer> WEBSOCKET_MAX_MESSAGE_LENGTH = new Key<>(5242880);
```
isbm: Can we use hexadecimals here? I think that `0x500000` looks/understood better. Because  `0x500000 / 0x400 / 0x400` equals 5MB. If 2 bytes per character (UTF16), then the limit is to 2,5MB.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
```diff
@@ -183,30 +199,63 @@ public void onOpen(Session session, EndpointConfig config) throws IOException {
...
+            throws MessageTooBigException {
+        if (last) {
+            String message;
+            if (partialMessageBuffer.length() == 0) {
+                message = partialMessage;
+            } else {
+                partialMessageBuffer.append(partialMessage);
+                message = partialMessageBuffer.toString();
+            }
+            if (maxMessageLength >= 0 && message.length() > maxMessageLength) {
```
isbm: Do we actually need `maxMessageLength >= 0` here?

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
```diff
@@ -183,30 +199,63 @@ public void onOpen(Session session, EndpointConfig config) throws IOException {
...
+                throw new MessageTooBigException(maxMessageLength);
+            }
+
+            // Notify all registered listeners
+            if (!message.equals("server received message")) {
+                // Salt API adds a "data: " prefix that we need to ignore
+                Event event = JsonParser.EVENTS.parse(message.substring(6));
+                synchronized (listeners) {
+                    listeners.stream().forEach(l -> l.notify(event));
```
isbm: It is up to you, but maybe we should not use one character variables. Just a nitpick.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
```diff
@@ -183,30 +199,63 @@ public void onOpen(Session session, EndpointConfig config) throws IOException {
...
+                }
+            }
+
+            // Reset the size to the defaultBufferSize and empty the buffer
+            partialMessageBuffer.setLength(defaultBufferSize);
+            partialMessageBuffer.trimToSize();
+            partialMessageBuffer.setLength(0);
+        } else {
+            partialMessageBuffer.append(partialMessage);
+            if (maxMessageLength >= 0 && partialMessageBuffer.length() > maxMessageLength) {
```
isbm: Same question here about first check for `0`.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
```diff
@@ -183,30 +199,63 @@ public void onOpen(Session session, EndpointConfig config) throws IOException {
...
-        this.close();
-        t.printStackTrace();
+    public void onError(Throwable t) throws IOException {
+        CloseReason closeReason;
+        if (t instanceof MessageTooBigException) {
+            closeReason = new CloseReason(CloseCodes.TOO_BIG, t.getMessage());
+        } else {
+            closeReason = new CloseReason(CloseCodes.CLOSED_ABNORMALLY, t.getMessage());
+        }
+        close(closeReason);
```
isbm:
```java
@OnError
public void onError(Throwable exc) throws IOException {
    close(new CloseReason(exc instance of MessageTooBigException 
                          ? CloseCodes.TOO_BIG : CloseCodes.CLOSED_ABNORMALLY, exc.getMessage));
}
```

:wink:

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
renner: It was wrong, I fixed it to `maxMessageLength > 0`: The implementation allows you to disable the limit by setting a value <= 0.

src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
```diff
@@ -35,6 +35,13 @@
     public static final Key<String> PROXY_PASSWORD = new Key<>();
 
     /**
+     * Maximum websocket message length in characters. The default value corresponds to
+     * 10 MB, assuming that one character takes up 2 bytes. This limit can be disabled
+     * by setting a value less than or equal to 0.
+     */
+    public static final Key<Integer> WEBSOCKET_MAX_MESSAGE_LENGTH = new Key<>(5242880);
```
isbm: I am still for using hexadecimals here. And the doc above is anyway confusing and seems to be incorrect. If your data is two-bytes, then the actual limit supposed to be doubled. That is, you allow now 5MB only per one byte.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
renner: Agreed, changed with 0f24c5df2fc93008dffa33b9dd898cff0d46c12c.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
```diff
@@ -183,30 +199,63 @@ public void onOpen(Session session, EndpointConfig config) throws IOException {
...
      * @param t The Throwable object received on the current error.
+     * @throws IOException in case of an error when closing the session
      */
     @OnError
-    public void onError(Throwable t) {
-        this.close();
-        t.printStackTrace();
+    public void onError(Throwable t) throws IOException {
+        CloseReason closeReason;
+        if (t instanceof MessageTooBigException) {
```
isbm: Unless you are OK that `t` can be also a subtype of `MessageTooBigException`, I would use:
```java
if (t.getClass() == MessageTooBigException.class) {...
```

isbm: Ant if you already changed variable `l` to `listener`, maybe we could also change `t` to `throwable` or something similar?

src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java
renner: My calculation is correct I guess, note that the limit is defined in **characters** :wink: Changed the notation to hexadecimal though with a8a5cb9382171b77f9a7d08519d59bcd10fecb24.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
renner: I don't think there is anything bad about using `instanceof` here. Subclasses of `MessageTooBigExceptions` do not exist at this moment, but if they existed I would most likely want them to end up with `CloseCodes.TOO_BIG` as well.

isbm:
LGTM :wink:

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
renner: Done (and renamed `t` to `throwable`) with 746bd688a493797de86e9dae1216c2d05de8d003.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
```diff
@@ -183,30 +199,58 @@ public void onOpen(Session session, EndpointConfig config) throws IOException {
...
+            if (!message.equals("server received message")) {
+                // Salt API adds a "data: " prefix that we need to ignore
+                Event event = JsonParser.EVENTS.parse(message.substring(6));
+                synchronized (listeners) {
+                    listeners.stream().forEach(listener -> listener.notify(event));
+                }
+            }
+        } else {
+            partialMessageBuffer.append(partialMessage);
+            if (maxMessageLength > 0 && partialMessageBuffer.length() > maxMessageLength) {
```
AdamMajer: Move check before append. This actually could be checked before the `if(last)` branch and de-duplicate the code.

renner: I might be wrong, but actually my impression is that it is better to have two checks here considering the two different cases:

1. `last` is true: we have a complete message (either the incoming string `partialMessage` or `buffer.toString`) -> we check the length on the `String` object.
2. `last` is false: we append to the buffer -> we check the length of the buffer (`StringBuilder`) without converting it to a `String` object.

If we wanted to perform this check only once, we would either need to convert the buffer to a string on every iteration, which results in a String object every time, only for the check! Or we would need to use the buffer on every partial message, which is unnecessary in most cases since for 'small' events (only one part, happens very often) we can just take the incoming string and that's our message.

isbm: Is the double check of `maxMessageLength` necessary? The code says "Pass only when there is a door and I fit into that door". Which means, you won't be able to pass the door _anyway_ if its size is zero, i.e. it is absent.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
```diff
@@ -183,30 +199,58 @@ public void onOpen(Session session, EndpointConfig config) throws IOException {
...
+            } else {
+                partialMessageBuffer.append(partialMessage);
+                message = partialMessageBuffer.toString();
+
+                // Reset the size to the defaultBufferSize and empty the buffer
+                partialMessageBuffer.setLength(defaultBufferSize);
+                partialMessageBuffer.trimToSize();
+                partialMessageBuffer.setLength(0);
+            }
+            if (maxMessageLength > 0 && message.length() > maxMessageLength) {
```
isbm: Same here. I don't think 'maxMessageLength > 0' is any necessary here. We want to say that the message is too big, i.e. bigger than allowed length. So the 'message.length() bigger than the maximum message length" is just enough.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
AdamMajer: I mean, just doing,

     if (maxMessageLength > 0 && partialMessageBuffer.length() + partialMessage.length() > maxMessageLength) { 
        throw ...
    }

isbm: It is still enough to do:
```java
if (partialMessageBuffer.length() + partialMessage.length() > masMessageLength) {
...
}
```

AdamMajer: No, you can't do that because maxMessageLength is special. The check for >0 needs to remain.

    maxMessageLength = config.get(ClientConfig.WEBSOCKET_MAX_MESSAGE_LENGTH);

if it is <= 0, then max length is to be ignored.

renner: @isbm: Not really. Note that setting `maxMessageLength` to a value `<= 0` is supposed to disable the limit. We never want to throw an exception in this case so there is no need to compare anything. Look at your code: if `maxMessageLength` is set to 0, it would throw an exception for every message with at least one character!

@AdamMajer: I see what you mean, will try to de-duplicate in that way.

isbm: @renner Ah, you mean `maxMessageLength = 0` _also_ disables the check entirely. Then OK.

renner: De-duplicated with bb939050600ba8d5df6bfdc39fe25e139e015a52.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
```diff
@@ -64,13 +72,11 @@ public EventStream() {
...
+     * @throws SaltStackException in case of an error during stream initialization
      */
-    public EventStream(ClientConfig config) {
-        try {
-            initializeStream(config);
-        } catch (SaltStackException e) {
-            e.printStackTrace();
-        }
+    public EventStream(ClientConfig config) throws SaltStackException {
+        maxMessageLength = config.get(ClientConfig.WEBSOCKET_MAX_MESSAGE_LENGTH);
```
AdamMajer: Going by @isbm 's comment earlier (ie. misunderstanding what this parameter represents), I think that maybe the <= 0 case could be handled here, when config is accessed to initialize maxMessageLength parameter.

    maxMessageLength = (config.get(...) > 0 ? config.get(...) : Integer.MAX_VALUE);

then later we don't have to test for > 0 case and just use it at face value.

This is just a suggestion, it's also OK as is.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
```diff
@@ -183,30 +199,57 @@ public void onOpen(Session session, EndpointConfig config) throws IOException {
...
-    public void onMessage(String message) {
-        if (message != null && !message.equals("server received message")) {
-            // Salt API adds a "data: " prefix that we need to ignore
-            Event event = JsonParser.EVENTS.parse(message.substring(6));
-            synchronized (listeners) {
-                listeners.stream().forEach(l -> l.notify(event));
+    public void onMessage(String partialMessage, boolean last)
+            throws MessageTooBigException {
+        if (maxMessageLength > 0 &&
+                messageBuffer.length() + partialMessage.length() > maxMessageLength) {
```
AdamMajer: I understand I suggested this construct, but since this is external data and assuming "bad things" happen, a better way to write this that avoids possibility of integer overflow (which could bypass this check) would be,

    partialMessage.length() > maxMessageLength - messageBuffer.length()

The only difference here should be checked exception vs. unchecked exception from messageBuffer in case of "bad things".

AdamMajer:
Aside from my completely pedantic comments, I do not see anything more that could be improved here. So merge away :+1:

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
isbm: @AdamMajer  Not really. The whole mystery of the `maxMessageLength = 0` or `-1` etc, that it means to disable the very message length check. Consequentially, your code would invalidate this feature. Also default value is taken properly here: https://github.com/SUSE/saltstack-netapi-client-java/blob/master-support-partial-websocket-messages/src/main/java/com/suse/saltstack/netapi/config/ClientConfig.java#L108

renner: @isbm: No, @AdamMajer's proposed fix actually allows to remove the `maxMessageLength > 0` check, while keeping the feature of disabling the limit by setting 0 or -1. Setting the ultimate message length limit to `Integer.MAX_VALUE` makes sense to me, implemented with d62f995546a256739e1b89afdfa4f6bd3a145136.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
renner: Thanks, fixed with 04da23704f97d0e2d85212399b3ec4fce6662c39.

src/main/java/com/suse/saltstack/netapi/event/EventStream.java
isbm: OK, this alters the very idea of disabling the check and _implicitly_ set to the integer maximum and now it makes it "Either less than maximum but more than zero, or maximum". This makes perfect sense.

renner:
I merged it, thanks a lot for all your feedback @AdamMajer and @isbm!

PR #115, opened 2015-09-21
Switch Pkg.Info to use ZonedTimeDate

lucidd:
This changes the `TimeDate` of `Pkg.Info` to include timezone information to reflect @isbm recent changes in salt.

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/calls/modules/Pkg.java
+++ b/src/main/java/com/suse/saltstack/netapi/calls/modules/Pkg.java
@@ -56,7 +56,7 @@ public class Pkg {
         private ZonedDateTime buildDate;
         @SerializedName("build_host")
         private String buildHost;
-        private String decription;
+        private String description;
         private String group;
         @SerializedName("install_date")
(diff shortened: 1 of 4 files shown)


isbm:
@lucidd Unit test? :chicken:

lucidd:
@isbm Works on my machine :stuck_out_tongue:  :poultry_leg:

isbm:
@lucidd Jenkins doesn't believe you. :-) Please add a unit test for this.

lucidd:
@isbm @renner i added a unit test for the `pkg.info_installed` call which includes parsing and checking the timezone. Also since this is the first unit test for the call API it can serve as example how to write unit tests for the rest.

isbm:
:kiss: 

As of me, merge it, @lucidd

AdamMajer:
Seems you can remove the LocalDateTime parser - it's not used anymore.

Looks good. :+1:

lucidd:
@AdamMajer i left it there because we will need it in the future since the plan is to long term move away from `Date` and use only the java 8 equivalents.

PR #116, opened 2015-09-22
implement Match module functions

lucidd:
This implements some of the functions in the match Module

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/calls/modules/Match.java
+++ b/src/main/java/com/suse/saltstack/netapi/calls/modules/Match.java
@@ -1,0 +1,60 @@
+package com.suse.saltstack.netapi.calls.modules;
+
+import com.google.gson.reflect.TypeToken;
+import com.suse.saltstack.netapi.calls.LocalCall;
+
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/**
+ * salt.modules.match
+ */
+public class Match {
+
+    public static LocalCall<Boolean> compound(String tgt, Optional<String> minionId) {
+        LinkedHashMap<String, Object> args = new LinkedHashMap<>();
+        args.put("tgt", tgt);
+        minionId.ifPresent(id -> args.put("minion_id", id));
+        return new LocalCall<>("match.compound", Optional.empty(),
+                Optional.of(args), new TypeToken<Boolean>(){});
+    }
+
+    public static LocalCall<Boolean> compound(String tgt) {
+        return compound(tgt, Optional.empty());
+    }
+
+    public static LocalCall<Boolean> glob(String tgt, Optional<String> minionId) {
+        LinkedHashMap<String, Object> args = new LinkedHashMap<>();
+        args.put("tgt", tgt);
+        minionId.ifPresent(id -> args.put("minion_id", id));
+        return new LocalCall<>("match.glob", Optional.empty(),
+                Optional.of(args), new TypeToken<Boolean>(){});
+    }
+
+    public static LocalCall<Boolean> glob(String tgt) {
+        return glob(tgt, Optional.empty());
+    }
+
+    public static LocalCall<Boolean> grain(String tgt, Optional<String> delimiter) {
+        LinkedHashMap<String, Object> args = new LinkedHashMap<>();
+        args.put("tgt", tgt);
+        delimiter.ifPresent(d -> args.put("delimiter", d));
+        return new LocalCall<>("match.grain", Optional.empty(),
+                Optional.of(args), new TypeToken<Boolean>(){});
+    }
+
+    public static LocalCall<Boolean> list(List<String> tgt, Optional<String> minionId) {
+        LinkedHashMap<String, Object> args = new LinkedHashMap<>();
+        args.put("tgt", tgt.stream().collect(Collectors.joining(",")));
+        minionId.ifPresent(id -> args.put("minion_id", id));
+        return new LocalCall<>("match.list", Optional.empty(),
+                Optional.of(args), new TypeToken<Boolean>(){});
+    }
+
+    public static LocalCall<Boolean> list(String... tgt) {
+        return list(Arrays.asList(tgt), Optional.empty());
+    }
+}


src/main/java/com/suse/saltstack/netapi/utils/ClientUtils.java
```diff
@@ -63,4 +64,24 @@ public static ParameterizedType parameterizedType(Type ownerType, Type rawType,
...
+    /**
+     * Concatenates a list of strings with a delimiter
+     *
+     * @param list list of strings
+     * @param delimiter delimiter string
+     * @return the concatenated string separated by the delimiter
+     */
+    public static String join(List<String> list, String delimiter) {
+        if (list.isEmpty()) {
+            return "";
```
isbm: 42! This is not nice.
Especially, if you're fun of lambdas. :wink: 

```java
    public static String join(List<String> list, String delimiter) {
        StringBuilder out = new StringBuilder();
        if (list != null && !list.isEmpty()) {
            out.append(list.get(0));
            list.subList(1, list.size()).stream().forEach((item) -> {
                out.append(delimiter).append(item);
            });
        }

        return out.toString();
    }
```

src/main/java/com/suse/saltstack/netapi/utils/ClientUtils.java
```diff
@@ -63,4 +64,24 @@ public static ParameterizedType parameterizedType(Type ownerType, Type rawType,
...
+        } else {
+            StringBuilder sb = new StringBuilder(list.get(0));
```
isbm: Two letter variables are not nice either.

src/main/java/com/suse/saltstack/netapi/utils/ClientUtils.java
```diff
@@ -63,4 +64,24 @@ public static ParameterizedType parameterizedType(Type ownerType, Type rawType,
...
+            for (int i = 1; i < list.size(); i++) {
+                sb.append(delimiter);
+                sb.append(list.get(i));
+            }
+            return sb.toString();
```
isbm: And two returns. :wink:

src/test/java/com/suse/saltstack/netapi/utils/ClientUtilsTest.java
```diff
@@ -64,4 +67,16 @@ public void testStreamToString() {
         assertEquals("Result doesn't match test string", TEST_STRING, result);
     }
 
+
+    @Test
+    public void testJoin() {
+        List<String> strings = Arrays.asList("Hello", "World", "123");
+        List<String> single = Arrays.asList("single");
+        List<String> empty = new ArrayList<>();
+
```
isbm:
```java
List<String> kaboom = null;
```

src/main/java/com/suse/saltstack/netapi/calls/modules/Match.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * salt.modules.match
+ */
+public class Match {
+
+    public static LocalCall<Boolean> compound(String tgt, Optional<String> minionId) {
```
isbm: What is `tgt` and where all javadocs were yanked?

lucidd: It does not make sense to document the module functions since they are already documented on the salt side. The variable names are borrowed from the documentation page as well to keep it consistent.  The only thing that may vary is the types if we can provide something saner on the java side.

src/test/java/com/suse/saltstack/netapi/utils/ClientUtilsTest.java
lucidd: If someone passes around `null` values then he deserves that the exception blows up in his face. :bomb:

src/main/java/com/suse/saltstack/netapi/calls/modules/Match.java
isbm: It makes sense to have a simple javadoc, what will popup in the IDE, so everyone can see it while programming. I don't know what means "tgt". Inability to write "target"? That Goofy Thing? Tyrese, Ginuwine and Tank R&B singers?..

src/test/java/com/suse/saltstack/netapi/utils/ClientUtilsTest.java
isbm: Fair enough. You may remove the null check in the rewritten sub above then.

src/main/java/com/suse/saltstack/netapi/calls/modules/Match.java
lucidd: Even if we provide documentation it would be mostly copied from the salt docs and would not be precise since our library may work with different versions of the module which could have slightly different but documented behavior.

isbm: Bindings that are not documented are crappy. It is not a port of Salt in Java that we could just refer to the original documentation. E.g. ZMQ has bindings [documented](http://zeromq.github.io/jzmq/javadocs/org/zeromq/ZMQ.Socket.html) although they do replicate the original docs.

isbm: You probably don't need to document it so much as in Salt, but at least you could say in one sentence what it is, what is return and what are params. I bet your IDE cries for it and checkstyle hates you for this. :wink:

src/main/java/com/suse/saltstack/netapi/calls/modules/Match.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/**
+ * salt.modules.match
+ */
+public class Match {
```
isbm: BTW, @renner suggests to have at least have a documentation for the whole class with the URL link to it. Maybe Johannes will write more here.

lucidd: I will open a separate issue to track the documentation of the modules where we also can discuss what to do.

renner: Thanks for opening the issue, @lucidd! From my perspective there is no need to add javadoc for the `Match` module class within this patch, especially I wouldn't want to copy it from the salt documentation for the reasons mentioned in #117. For now let's just do it as the other module classes we already have, i.e. keep it as it is.

src/test/java/com/suse/saltstack/netapi/utils/ClientUtilsTest.java
```diff
@@ -63,5 +63,4 @@ public void testStreamToString() {
                 new ByteArrayInputStream(TEST_STRING.getBytes()));
         assertEquals("Result doesn't match test string", TEST_STRING, result);
     }
-
```
renner: While this change makes sense it does not actually seem to belong to this patch.

lucidd: Oops seems like a left over from a previous version i will remove it. (or rather add it back :smile: )

PR #119, opened 2015-09-24
implement event parsers to help dealing with events

lucidd:
implement event parsers to help dealing with events

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/event/BeaconEvent.java
+++ b/src/main/java/com/suse/saltstack/netapi/event/BeaconEvent.java
@@ -1,0 +1,88 @@
+package com.suse.saltstack.netapi.event;
+
+import com.suse.saltstack.netapi.datatypes.Event;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Represents an event fired by beacons
+ */
+public class BeaconEvent {
+    private static final Pattern PATTERN =
+            Pattern.compile("^salt/beacon/([^/]+)/([^/]+)/(.*)$");
+
+    private final String beacon;
+    private final String minionId;
+    private final String additional;
+    private final Map<String, Object> data;
+
+    /**
+     * Creates a new BeaconEvent
+     * @param minionId the id of the minion sending the event
+     * @param beacon the beacon name
+     * @param additional additional information depending on the beacon
+     * @param data data containing more information about this event
+     */
+    public BeaconEvent(String minionId, String beacon, String additional,
+            Map<String, Object> data) {
+        this.minionId = minionId;
+        this.beacon = beacon;
+        this.additional = additional;
+        this.data = data;
+    }
+
+    /**
+     * Returns the beacon name.
+     *
+     * @return the beacon name
+     */
+    public String getBeacon() {
+        return beacon;
+    }
+
+    /**
+     * Returns the id of the minion that triggered the beacon
+     *
+     * @return the minion id
+     */
+    public String getMinionId() {
+        return minionId;
+    }
+
+    /**
+     * Provides additional information from the tag depending on the type of beacon
+     *
+     * @return additional information
+     */
+    public String getAdditional() {
+        return additional;
+    }
+
+    /**
+     * The event data containing more information about this event
+     *
+     * @return the event data
+     */
+    public Map<String, Object> getData() {
+        return data;
+    }
+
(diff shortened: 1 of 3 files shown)


src/main/java/com/suse/saltstack/netapi/event/BeaconEvent.java
```diff
@@ -0,0 +1,46 @@
...
+        return minionId;
+    }
+
+    public Map<String, Object> getData() {
+        return data;
+    }
+
+    public static Optional<BeaconEvent> parse(Event event) {
+        Matcher m = ret.matcher(event.getTag());
+        if(m.matches()) {
```
isbm: @cyntss will love this. :wink: (hint: spaces)

src/main/java/com/suse/saltstack/netapi/event/BeaconEvent.java
```diff
@@ -0,0 +1,46 @@
...
+    public String getMinionId() {
+        return minionId;
+    }
+
+    public Map<String, Object> getData() {
+        return data;
+    }
+
+    public static Optional<BeaconEvent> parse(Event event) {
+        Matcher m = ret.matcher(event.getTag());
```
isbm: Please, let us stop one-character variables in the code once and for ever. This is not nice.

src/main/java/com/suse/saltstack/netapi/event/MinionStartEvent.java
```diff
@@ -0,0 +1,39 @@
...
+        return minionId;
+    }
+
+    public Map<String, Object> getData() {
+        return data;
+    }
+
+    public static Optional<MinionStartEvent> parse(Event event) {
+        Matcher m = ret.matcher(event.getTag());
+        if(m.matches()) {
```
isbm: Spaces + 1 char var.

src/main/java/com/suse/saltstack/netapi/event/ResultEvent.java
```diff
@@ -0,0 +1,46 @@
...
+        return minionId;
+    }
+
+    public Map<String, Object> getData() {
+        return data;
+    }
+
+    public static Optional<ResultEvent> parse(Event event) {
+        Matcher m = ret.matcher(event.getTag());
+        if(m.matches()) {
```
isbm: Spaces + 1 char var

src/main/java/com/suse/saltstack/netapi/event/BeaconEvent.java
AdamMajer: What is an acceptable minimum? As an example, which one of following would be acceptable? (I guess option 1 is there for posterity)

    1. for (int i = 0; i < pos; ++i)
    2. for (int idx = 0; idx < pos; ++idx)
    3. for (int index = 0; index < position; ++index)

lucidd:
Guys this PR is obviously not done yet (hence draft in the title) i created it to get some feedback on the design not the style complains checkstyle already tells me about :stuck_out_tongue_winking_eye:

src/main/java/com/suse/saltstack/netapi/event/BeaconEvent.java
isbm: @AdamMajer I would say, this example is a bit out of the context, because you use integer temporary variables, that are actually allowed for historical reasons and used as an [exceptions from earlier times](https://www.cwu.edu/~gellenbe/javastyle/variable.html).

Still, as per [Oracle Java Code Conventions](http://www.oracle.com/technetwork/java/codeconventions-135099.html), _"one-character variable names should be **avoided** except for temporary "throwaway" variables. Common names for temporary variables are `i`, `j`, `k`, `m`, and `n` for **integers**; `c`, `d`, and `e` for **characters**"_. The Matcher instance is nor an integer neither character.

Also Google [avoiding](https://google.github.io/styleguide/javaguide.html#s5.2.7-local-variable-names) all one-character local variables.

Also in the Salt itself rules are at least three characters for _everything_, otherwise PyLint will just refuse your code.

So I think we should stick to that too. So @lucidd, please fix that. :wink:

isbm:
@lucidd ha!, but didn't @renner asked you to provide better descriptions? :tongue:

src/main/java/com/suse/saltstack/netapi/event/BeaconEvent.java
AdamMajer: @isbm thank you for a clear explanation,

Issue #99, opened 2015-08-24
Refactor the main class


renner:
The main library class (`SaltStackClient.java`) has grown quite a bit, especially since the recent merge of #94. It could make sense to refactor this class now, moving away the new methods (`callSync()` and `callAsync()`) into a separate *module* of the library for instance.

AdamMajer:
Can you clarify term *module*?

The `callSync()` and `callAsync()` members could be moved as member functions of `calls` package taking `SaltStackClient` as parameter. So for example, `WheelCall.java` would have a `callSync(SaltStackClient client)` and similar async function.

renner:
Sorry for the late reply, I mean a separate class, while leaving it open how this should be designed. What would be important to me is that we create a nice API for the users of the library that allows to write beautiful and readable code. Thanks for your pull request, I will have a look and give feedback there.

renner:
Closing this issue after the merge of #120.

PR #120, opened 2015-09-25
Move sync/async remote call setups to their respective classes

AdamMajer:
There is no logic changes. The only major change is that the generic
SaltStackClient.call() function is now exposed as a public function.

Issue: #99 

As always, comments are welcome.


isbm:
Looks OK

lucidd:
The only downside of this approach i can see is that before you could just call `callSync/callAsync` and it did not matter if you gave it a runner function a wheel function or a local function (with the caveat that you had to provide a additional target). This could also be positive since its explicit now but i'm not completely sure.

renner:
I merged this patch, thank you @AdamMajer and sorry for the really long delay!

PR #123, opened 2015-09-27
Merge ResultInfo into Jobs.Info<R>

AdamMajer:
ResultInfo is parsing the same structure as Jobs.Info. There is no
reason keeping both around.


renner:
Hey @AdamMajer, we are eventually closing this PR in favour of a different implementation that takes into account the most recent changes regarding event parsing. Thanks for your efforts anyways!

Issue #75, opened 2015-06-02
Use TypeAdapter over JsonDeserializer


renner:
[Gson Javadoc for `JsonDeserializer`] (https://google-gson.googlecode.com/svn/trunk/gson/docs/javadocs/com/google/gson/JsonDeserializer.html) says: "New applications should prefer TypeAdapter, whose streaming API is more efficient than this interface's tree API". We still have some implementations of `JsonDeserializer` in the `JsonParser` class that should be converted, namely:

- `StatsDeserializer`
- `ArgumentsDeserializer`

hustodemon:
(We should bear in mind that `JsonDeserializer` is null-safe by default (according to my research it uses null-safe `TreeTypeAdapter` internally).)

wallner:
I would like to try my hand on this, is this still being worked on? If not I would like to have a got at it this week.

lucidd:
@wallner I'm not aware of anyone actively working on this issue so feel free to take it on. PRs are always welcome :smile:

renner:
There is no `JsonDeserializer` implementations left after the merge of #124. Thanks for the patch @mathall! Closing this issue.

PR #124, opened 2015-10-05
Replace JsonDeserializer implementations with TypeAdapter

mathall:
This removes the last usage of JsonDeserializer and solves issue #75

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
+++ b/src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
@@ -84,7 +84,7 @@ public class JsonParser<T> {
                 .registerTypeAdapter(LocalDateTime.class, new LocalDateTimeISOAdapter())
                 .registerTypeAdapter(ZonedDateTime.class, new ZonedDateTimeISOAdapter())
                 .registerTypeAdapter(StartTime.class, new StartTimeAdapter().nullSafe())
-                .registerTypeAdapter(Stats.class, new StatsDeserializer())
+                .registerTypeAdapter(Stats.class, new StatsAdapter())
                 .registerTypeAdapter(Arguments.class, new ArgumentsDeserializer())
                 .registerTypeAdapterFactory(new OptionalTypeAdapterFactory())
                 .create();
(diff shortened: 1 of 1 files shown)


src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -84,8 +81,8 @@ public JsonParser(TypeToken<T> type) {
                 .registerTypeAdapter(LocalDateTime.class, new LocalDateTimeISOAdapter())
                 .registerTypeAdapter(ZonedDateTime.class, new ZonedDateTimeISOAdapter())
                 .registerTypeAdapter(StartTime.class, new StartTimeAdapter().nullSafe())
-                .registerTypeAdapter(Stats.class, new StatsDeserializer())
-                .registerTypeAdapter(Arguments.class, new ArgumentsDeserializer())
+                .registerTypeAdapter(Stats.class, new StatsAdapter().nullSafe())
```
lucidd: i don't think we need the `nullSafe` here. If something can be `null` is should generally marked with `Optional` then the `Optional` parser will take care of the null value.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -176,118 +173,108 @@ public void write(JsonWriter out, Optional<A> optional) throws IOException {
...
+        public void write(JsonWriter jsonWriter, Arguments args) throws IOException {
+            throw new UnsupportedOperationException("Writing JSON not supported.");
+        }
 
-            if (json != null && json.isJsonArray()) {
-                for (JsonElement jsonElement : json.getAsJsonArray()) {
-                    fillArgs(result, jsonElement);
+        @Override
+        public Arguments read(JsonReader jsonReader) throws IOException {
+            if (jsonReader.peek() == JsonToken.NULL) {
```
lucidd: Also here the `null` checks should make it fail similar to `LocalDateTimeISOAdapter`.

src/main/java/com/suse/saltstack/netapi/parser/JsonParser.java
```diff
@@ -176,118 +173,107 @@ public void write(JsonWriter out, Optional<A> optional) throws IOException {
...
+                throw new JsonParseException("null is not a valid value for Arguments");
+            } else if (jsonReader.peek() != JsonToken.BEGIN_ARRAY) {
```
mathall: I reconsidered this check, I think to maintain the old behaviour we should consume whatever is next in the stream and then return null. Is this the desired behaviour though? Whether we keep it as it is now, throw an exception like with the null check, or just continue without the check all together, the outcome will likely be the same; parse error. Unless you strongly want to keep it as it was, I'd say I might as well just remove this check and let beingArray() fail.

lucidd: Yeah that check does not make much sense feel free to remove it.

mathall: Done. Please note that the behaviour is now slightly changed with this PR. A non-array Arguments-value will now cause parse error where as before it would silently let it slide, leaving the deserialized Arguments empty. (a change for the better I'd say)

lucidd:
LGTM :+1:

renner:
Merged, thanks for your contribution!

PR #127, opened 2015-10-15
Allow adding event listeners before stream initialization

renner:
This patch allows to add event listeners to the websocket based event stream before the stream will actually be initialized. This way we can be sure that we won't miss any events between the handshake and adding the listener, which especially helps to make the unit tests more stable (see `TyrusWebSocketEventsTest`).


lucidd:
LGTM :+1:

PR #129, opened 2015-10-22
Tests for `cmd' and `test' module

mbologna:
I added some JUnit tests for `cmd' and `'test' module.


src/test/java/com/suse/saltstack/netapi/calls/modules/TestTest.java
```diff
@@ -0,0 +1,48 @@
...
+                .getResourceAsStream("/modules/test/versions_information.json");
+        JsonParser<VersionInformation> parser = new JsonParser<>(type);
+        VersionInformation parsed = parser.parse(is);
+        assertEquals("2015.8.0-574-g99384bc", parsed.getSalt().get("Salt"));
+        assertEquals("3.13.0-65-generic", parsed.getSystem().get("release"));
+        assertEquals("Ubuntu 14.04 trusty", parsed.getSystem().get("dist"));
+        assertEquals("Ubuntu 14.04 trusty", parsed.getSystem().get("system"));
+        assertNull(parsed.getDependencies().get("cherrypy"));
+        assertEquals("2.7.6 (default, Jun 22 2015, 17:58:13)",
+                parsed.getDependencies().get("Python"));
```
lucidd: I think it would be good to at least assert that `Dependency Versions` has the right size if we don't test all the entries.

src/test/java/com/suse/saltstack/netapi/calls/modules/TestTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Test unit tests.
+ */
+public class TestTest {
+
+    @Test
+    public void testTestPing() {
+        TypeToken<Boolean> type =
+                com.suse.saltstack.netapi.calls.modules.Test.ping().getReturnType();
```
lucidd: I think we should use the fully qualified name for the `@Test` since nobody looks at them anyway so it does not matter if they are longer but making the test code shorter should make it more readable.

src/test/java/com/suse/saltstack/netapi/calls/modules/TestTest.java
mbologna: Yep!

src/test/java/com/suse/saltstack/netapi/calls/modules/TestTest.java
mbologna: Sure, code readability is better when inverting imports. Done in the next commit!

PR #131, opened 2015-10-22
Add support for `pkg' module methods and tests

mbologna:
`pkg' module is a proxy for the underlining 
distro of choice where SaltStack is installed.
So not every package manager supports every 
operation.

In this case, I have used SaltStack develop (from
git) and aptpkg frontend.

Add support for:
- upgrade_available, which returns whether a
package has an upgrade available from distro
repo
- latest_version, which return the latest version 
of the named package available for upgrade.

CAVEAT: if more than one package name is 
specified, a dict of name/version 
pairs is returned.

This differentiates the implementation when we 
call this method with one argument or with 
varargs arguments.


src/main/java/com/suse/saltstack/netapi/calls/modules/Pkg.java
```diff
@@ -188,4 +188,32 @@ private Pkg() { }
...
+    }
+
+    public static LocalCall<String> latestVersion(String packageName) {
+        LinkedHashMap<String, Object> args = new LinkedHashMap<>();
+        args.put("package", packageName);
+        return new LocalCall<>("pkg.latest_version", Optional.empty(), Optional.of(args),
+                new TypeToken<String>(){});
+    }
+
+    public static LocalCall<Map<String, String>> latestVersions(String... packages)
```
lucidd: Lets avoid the exception by making the requirements explicit and take at least 2 package names.

```java
latestVersions(String first, String second, String... packages)
```

this function can be only called with a minimum 2 package names.

lucidd:
In general i like the patch but we need to do some more research first because when i execute `pkg.latest_version` on a suse machine i get back something like this.

`salt 'minion2' pkg.latest_version vim --out json`
```
{
    "minion2": "7.4.326-2.62"
}
```

`salt 'minion2' pkg.latest_version emacs vim --out json`
```
{
    "minion2": {
        "emacs": {
            "version": "24.3-14.44", 
            "vendor": "SUSE LLC <https://www.suse.com/>"
        }, 
        "vim": {
            "version": "7.4.326-2.62", 
            "vendor": "SUSE LLC <https://www.suse.com/>"
        }
    }
}
```

So the we need to first figure out what kind of return formats are out there and why. Maybe this is even something salt needs to fix (unify).

mbologna:
@lucidd it returns a string if only a package name is supplied, otherwise returns a map - should be `{ name, version }` as written on documentation.
My .json was using `salt.modules.aptpkg` as frontend, while yours is using `salt.modules.zypper`.
In the end I think it's as you wrote, it depends on the particular implementation of pkg related to the considered distro.

src/main/java/com/suse/saltstack/netapi/calls/modules/Pkg.java
mbologna: Brilliant, I like your idea. Inserted in in the next commit!

src/main/java/com/suse/saltstack/netapi/calls/modules/Pkg.java
```diff
@@ -188,4 +188,26 @@ private Pkg() { }
...
+                new TypeToken<String>(){});
+    }
+
+    public static LocalCall<Map<String, String>> latestVersions(String firstPackageName,
+            String secondPackageName, String...packages) {
+        return new LocalCall<>("pkg.latest_version",
+                    Optional.of(Arrays.asList(firstPackageName,
+                            secondPackageName, packages)),
+                Optional.empty(), new TypeToken<Map<String, String>>(){});
+    }
```
renner: It is necessary to have those three parameters here in order to make sure that there is at least two package names were actually given, right? That would make sense in case Salt decides on the return type based on the incoming parameters. I would suggest to still use the same name as the name of the Salt module though (`latestVersion()`) just to make clear that it maps to the same module. Apart from that indentation looks a but weird here, maybe it could be improved a bit?

src/main/java/com/suse/saltstack/netapi/calls/modules/Pkg.java
```diff
@@ -188,4 +188,26 @@ private Pkg() { }
...
+        args.put("package", packageName);
+        return new LocalCall<>("pkg.upgrade_available", Optional.empty(), Optional.of(args),
+                new TypeToken<Boolean>(){});
+    }
+
+    public static LocalCall<String> latestVersion(String packageName) {
+        LinkedHashMap<String, Object> args = new LinkedHashMap<>();
+        args.put("package", packageName);
+        return new LocalCall<>("pkg.latest_version", Optional.empty(), Optional.of(args),
+                new TypeToken<String>(){});
```
renner: The [documentation] (https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.zypper.html#salt.modules.zypper.latest_version) doesn't actually mention that there is a `package` keyword argument. Did that change since this PR was opened or is it just not correctly documented?

renner:
Hey @mbologna, I would like to finally merge this for the next release, sorry for the long waiting time! Could you please either fix the conflicts with the current master, or redo the pull request on a branch of the SUSE project so I can fix the conflicts myself? Thanks! In case there is differences in the underlying modules behavior (e.g. `aptpkg` vs. `zypper`) I would recommend to figure out what is the expected **common** behaviour, implement that and open issues in upstream accordingly.

src/main/java/com/suse/saltstack/netapi/calls/modules/Pkg.java
mbologna: Unified the two methods name with `latestVersion()` and fixed indentation.
We could also unify them in a unique method, using varargs. But then, inside the method, we should differentiate the return type based on the cardinality of the argument (`packageName`).
What do you think?

src/main/java/com/suse/saltstack/netapi/calls/modules/Pkg.java
mbologna: Removed erroneous `package` kw.

mbologna:
Thanks @renner ! 
I pushed some commits that should resolve the issues.

renner:
Merged it, thanks for those fixes!

PR #134, opened 2015-11-12
Fix run json

### LINKED ISSUE #126: Saltstack version compatibility?
Hi guys,

I am running into a problem when trying to use this module. I am trying to issue an ad-hoc command to a minion via the run() method in the SaltStackClient class. Here is the code I am trying:
```
SaltStackClient saltClient = new SaltStackClient("http://salt-server");
ResultInfo resultInfo = saltClient.run(
        "user",
        "pass",
        AuthModule.PAM,
        "local",
        new MinionList("myhostname"),
        "cmd.run",
        Collections.singletonList("pwd"),
        Collections.emptyMap());
```
The above code produces a NullPointerException when trying to do ```resultInfo.getResult("myhostname")``` to get the output. I did some debugging and grabbed the request body that gets generated and used Postman to verify that the server accepts it and sends a response. Here is an example:
```
{
  "return": [
    {
      "myhostname": "/root"
    }
  ]
}
```
I think there is a mismatch between the structure of this response and what the json parser is expecting. Can you confirm this suspicion? I am using version 0.6.0 of this library and tried with Salt versions 2015.8.0 and 2015.5.2.

mbologna:
Fixed parsing for `cmd.run` (new versions of Saltstack should only return "A dictionary with the result 
of the execution, keyed by minion ID. A compound command will return a sub-dictionary 
keyed by function name.").
Adjusting parser accordingly.

Fix #126

### DIFF
--- a/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
+++ b/src/main/java/com/suse/saltstack/netapi/client/SaltStackClient.java
@@ -1,5 +1,19 @@
 package com.suse.saltstack.netapi.client;
 
+import static com.suse.saltstack.netapi.utils.ClientUtils.parameterizedType;
+
+import java.lang.reflect.Type;
+import java.net.URI;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
 import com.google.gson.Gson;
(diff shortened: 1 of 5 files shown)


renner:
I tested and merged your patch, thanks for the fix! One side note though: please do not re-order imports in a patch that is actually about something else. Your submitted patch should contain only relevant changes, while re-ordering imports would need to be done separately (if needed). I added 960d9aa58f41ff4a3d8889389567ddea825cda35 to fix the order (will add a checkstyle rule for this) and 14b851bde595a52790130bbb0bf7c3088e7b63c8 to renaming the new JSON file.

mbologna:
Wops! Sorry about import reordering. Thank you!

PR #135, opened 2015-11-16
Add example code

renner:
This pull request is about providing some code examples within the project to be used as a reference for people who like to get started with the library.

### DIFF
--- a/src/test/java/com/suse/saltstack/netapi/examples/Calls.java
+++ b/src/test/java/com/suse/saltstack/netapi/examples/Calls.java
@@ -1,0 +1,63 @@
+package com.suse.saltstack.netapi.examples;
+
+import com.suse.saltstack.netapi.AuthModule;
+import com.suse.saltstack.netapi.calls.WheelResult;
+import com.suse.saltstack.netapi.calls.modules.Grains;
+import com.suse.saltstack.netapi.calls.modules.Test;
+import com.suse.saltstack.netapi.calls.wheel.Key;
+import com.suse.saltstack.netapi.client.SaltStackClient;
+import com.suse.saltstack.netapi.datatypes.target.Glob;
+import com.suse.saltstack.netapi.datatypes.target.MinionList;
+import com.suse.saltstack.netapi.datatypes.target.Target;
+import com.suse.saltstack.netapi.exception.SaltStackException;
+
+import java.net.URI;
+import java.util.List;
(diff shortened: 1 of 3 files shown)


src/test/java/com/suse/saltstack/netapi/examples/Calls.java
```diff
@@ -0,0 +1,63 @@
...
+        // Init the client
+        SaltStackClient client = new SaltStackClient(URI.create(SALT_API_URL));
+
+        // Ping all minions using a glob matcher
+        Target<String> globTarget = new Glob("*");
+        Map<String, Boolean> results = client.callSync(
+                Test.ping(), globTarget, USER, PASSWORD, AuthModule.AUTO);
+
+        System.out.println("--> Ping results:\n");
+        results.keySet().forEach(
```
hustodemon: This could be written in more compact way:
```java
results.forEach((key, val) -> System.out.println(key + " -> " + val));
```
I found this "map shorthand" out recently and think it's a bit nicer.

src/test/java/com/suse/saltstack/netapi/examples/Calls.java
```diff
@@ -0,0 +1,63 @@
...
+                key -> System.out.println(key + " -> " + results.get(key)));
+
+        // Get the grains from a list of minions
+        Target<List<String>> minionList = new MinionList("minion1", "minion2");
+        Map<String, Map<String, Object>> grains = client.callSync(
+                Grains.items(true), minionList, USER, PASSWORD, AuthModule.AUTO);
+
+        for (String minion : grains.keySet()) {
```
hustodemon: Also here, it could be shortened.
```java
grains.forEach((minion, grain) -> {
    System.out.println("\n--> Listing grains for '" + minion + "':\n");
    grain.forEach((key, val) -> System.out.println(key + ": " + val));
});
```
Just a tip :smile_cat:

hustodemon:
Appart from those minor comments, it looks good to me :+1:

src/test/java/com/suse/saltstack/netapi/examples/Calls.java
renner: Thanks for the tips, implemented with 65e3726.

renner:
Thanks for the review, @hustodemon! I am going to merge this now, we can still add more example code later.

PR #137, opened 2015-12-04
Add support for Gradle builds

mbologna:
Added support for building the project using Gradle, a replacement
for maven with a plain-text config file.
Added Gradle cache files to .gitignore.

Tested with Gradle 2.9.


moio:
Hey @mbologna!

Thanks for this contribution as usual. I will leave it up to my colleagues to review it and to decide whether to merge it, I am chiming in now just to let you know of a particular complication we will run into if this is accepted. That complication arises from the way we distribute software in SUSE in general - and that also applies to this library.

As you might know openSUSE, SLE and all of their variant distros are rooted in the [RPM](http://www.rpm.org/max-rpm/ch-intro-to-rpm.html) packaging format and related tools, notably the [Open Build Service](http://openbuildservice.org/). That basically means we never ship software in JAR or WAR format - we always have to "wrap" those in RPM packages for trackability, reproducibility, security and other reasons.

Now for Java software, and Java libraries in the SUSE Manager context in particular, we make use of a tool, [tetra](https://github.com/moio/tetra), to ease the creation of the [specfiles](http://www.rpm.org/max-rpm/ch-rpm-basics.html) - descriptors used to generate those RPMs. JFYI: in general packaging Java software in RPM format properly can be very difficult, tetra takes a couple of shortcuts to make the process manageable in hours/days instead of months.

Now the pain point is that tetra currently supports ant and maven - gradle builds are only "supported manually" and that takes more work. If you ever want to dive into this topic and need more information I will be glad to answer any question!

Happy hacking!

renner:
Hello @mbologna! I generally like the non-xml configuration with gradle, but unfortunately - because of the reasons mentioned by @moio - we currently cannot migrate this project to a build system other than ant or maven. Since we also don't want to maintain support for two different build systems at the same time I would ask you if you are OK with closing this PR, at least until gradle is fully supported by tetra? Thanks for your efforts anyway!

mbologna:
Hi @moio !
Thank you for the clear under-the-hood explanation of why gradle can't be used with tetra. 

I will have a look at tetra and decide how to go from there.

mbologna:
Hi @renner !
No problem in closing this PR for the same reason you mention (specifically, using two build system).
Thanks!

PR #143, opened 2016-02-15
salt.modules.test.rand_str

jakubgasiorski:
Return a random string.


jakubgasiorski:
Please let me know if anything needs improvements.

lucidd:
Hi first of all thanks for your contribution. There are two things that need to be improved. The type parameter `R` in `LocalCall<R>` represents the type of one individual result from a minion not the full result of the call. For example for `test.ping` returns a `Boolean` for each minion so the the type is `LocalCall<Boolean>`. In your case for `rand_str` each minion returns a `String` to the type would be `LocalCall<String>` instead of `LocalCall<Map<String, List<Map<String, String>>>>`.

The second improvement would be to add support for the parameters of [rand_str](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.test.html#salt.modules.test.rand_str).Since they are both not required they should be made `Optional`. You can see an example of how `Optional` parameters can be handled [here](https://github.com/SUSE/salt-netapi-client/blob/master/src/main/java/com/suse/salt/netapi/calls/modules/Match.java#L18-L24).

jakubgasiorski:
Hi, 

thank you for detailed explanation. Hope next commit is better. However if anything need to be fixed, please let me know. 

Best regards, 
Jakub

src/main/java/com/suse/salt/netapi/calls/modules/Test.java
```diff
@@ -151,4 +152,13 @@ public VersionInformation(Map<String, String> salt,
         return new LocalCall<>("test.provider", Optional.empty(), Optional.of(args),
                 new TypeToken<String>() {});
     }
+
+    public static LocalCall<String> randStr(Optional<String> size,
```
lucidd: size should be an `Optional<Integer>`.

src/main/java/com/suse/salt/netapi/calls/modules/Test.java
```diff
@@ -31,6 +31,7 @@
             new LocalCall<>("test.providers", Optional.empty(), Optional.empty(),
             new TypeToken<Map<String, String>>(){});
 
+
```
lucidd: random new line :smile:

jakubgasiorski:
Done :) In next steps I would take more from test module.

lucidd:
Now it should work. There is one more enhancement i can think of that would be nice. Since the `hash_type` only accepts certain values like `md5` `sha1` etc it would be nice if it would take an enum instead of the String where the enum contains all the valid values you can pass.

jakubgasiorski:
Ineed that's quite good idea. I will add it in next step. Is there any common place for enums in package?

lucidd:
we usually put all custom types used in a certain module in the same file since there is usually no overlap between modules.

lucidd:
Awesome :+1:  merging it via https://github.com/SUSE/salt-netapi-client/commit/7cfe3740e508728b28418dedfe3df1b8d187035c

PR #144, opened 2016-03-09
Add support for state.show_highstate

renner:
This patch adds support for [`state.show_highstate`] (https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.state.html#salt.modules.state.show_highstate), returning just an `Object` right now.


renner:
Closing this as there is now a duplicate as #153.

PR #148, opened 2016-04-22
Better event parsing

lucidd:
this PR is smaller then it looks it depends on https://github.com/SUSE/salt-netapi-client/pull/146 and only adds one commit to it for better event handling. The new event parsing allows for staged parsing so inner datatypes can be chosen on information in the outer type.


renner:
@lucidd: Can't you make a PR against the #146 branch then? Duplicating all the changes is not really needed, but makes the review kind of difficult.

PR #152, opened 2016-05-02
Add target for specifying minions by compound pattern

akanto:
Ability to target minions by using compound matcher: https://docs.saltstack.com/en/latest/topics/targeting/compound.html

### DIFF
--- a/src/main/java/com/suse/salt/netapi/datatypes/target/Compound.java
+++ b/src/main/java/com/suse/salt/netapi/datatypes/target/Compound.java
@@ -1,0 +1,28 @@
+package com.suse.salt.netapi.datatypes.target;
+
+/**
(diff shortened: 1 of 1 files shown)


renner:
Hey @akanto, thanks for your contribution! Looking at the proposed patch, the only thing that makes the `Compound` class represent a compound matcher is the type string returned by `getType()`. So in order to cover more use cases with this, why don't we rather make a completely generic target class that takes a string as well as a matcher type (as an `enum` value)? It could be used then to represent **any** type of matcher, wouldn't that be much more useful in general?

renner:
@akanto: I am closing this because you were not responding in a while and the patch is not really useful in its current state. Feel free to reopen or make a new PR at any time. Thanks for your efforts anyways!

PR #153, opened 2016-05-04
state.apply with pillar data and state.show_highstate

mateiw:
* add `state.apply` call that takes pillar data
* add `state.show_highstate` call

### DIFF
--- a/src/main/java/com/suse/salt/netapi/calls/modules/State.java
+++ b/src/main/java/com/suse/salt/netapi/calls/modules/State.java
@@ -27,7 +27,7 @@ public class State {
 
     public static LocalCall<Map<String, Object>> apply(List<String> mods,
             Optional<Map<String, Object>> pillar) {
-        Map<String, Object> args = new LinkedHashMap();
+        Map<String, Object> args = new LinkedHashMap<>();
         args.put("mods", mods);
         if (pillar.isPresent()) {
             args.put("pillar", pillar.get());
--- a/src/main/java/com/suse/salt/netapi/calls/modules/State.java
+++ b/src/main/java/com/suse/salt/netapi/calls/modules/State.java
@@ -29,12 +29,8 @@ public class State {
             Optional<Map<String, Object>> pillar, Optional<Boolean> queue) {
         Map<String, Object> kwargs = new LinkedHashMap<>();
         kwargs.put("mods", mods);
-        if (pillar.isPresent()) {
-            kwargs.put("pillar", pillar.get());
-        }
-        if (queue.isPresent()) {
-            kwargs.put("queue", queue.get());
-        }
+        pillar.ifPresent(p -> kwargs.put("pillar", p));
+        queue.ifPresent(q -> kwargs.put("queue", q));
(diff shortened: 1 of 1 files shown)


src/main/java/com/suse/salt/netapi/calls/modules/State.java
```diff
@@ -28,4 +29,18 @@ private State() { }
...
     }
 
+    public static LocalCall<Map<String, Object>> apply(String mod,
+            Map<String, Object> pillar) {
+        LinkedHashMap args = new LinkedHashMap();
+        args.put("mods", Collections.singletonList(mod));
+        args.put("pillar", pillar);
+        return new LocalCall("state.apply", Optional.empty(), Optional.of(args),
+                new TypeToken<Map<String, Object>>() { });
+    }
```
renner: This method currently supports only a single state module, is that right? Why not just adding the `pillar` parameter to the existing apply method so you could give in a list of state modules? Then if we want we can have another method (without the `pillar` parameter) that could just call the other one, if you know what I mean.

renner:
I actually already created a PR with `state.show_highstate` (#144). We can close it and keep this one or you can also just add commits in #144, as you prefer.

src/main/java/com/suse/salt/netapi/calls/modules/State.java
```diff
@@ -18,14 +18,27 @@
...
+
+    public static LocalCall<Map<String, Object>> apply(List<String> mods,
+            Optional<Map<String, Object>> pillar) {
+        LinkedHashMap args = new LinkedHashMap();
         args.put("mods", mods);
-        return new LocalCall<>("state.apply", Optional.empty(), Optional.of(args),
+        if (pillar.isPresent()) {
+            args.put("pillar", pillar.get());
+        }
+        return new LocalCall("state.apply", Optional.empty(), Optional.of(args),
```
renner: Please fix warnings throughout this code:

1. the `LinkedHashMap` should be created with type arguments: `<String, Object>`?
2. the returned `LocalCall` should be parameterized, diamond operator?

renner:
Let's merge this one, @mateiw, @lucidd?

src/main/java/com/suse/salt/netapi/calls/modules/State.java
```diff
@@ -18,14 +18,30 @@
...
     public static LocalCall<Map<String, Object>> apply(String... mods) {
-        return apply(Arrays.asList(mods));
+        return apply(Arrays.asList(mods), Optional.empty(), Optional.empty());
+    }
+
+    public static LocalCall<Map<String, Object>> apply(List<String> mods,
+            Optional<Map<String, Object>> pillar, Optional<Boolean> queue) {
+        Map<String, Object> kwargs = new LinkedHashMap<>();
+        kwargs.put("mods", mods);
+        if (pillar.isPresent()) {
```
lucidd: use `ifPresent` for this and the one below to avoid `get`

renner: ok, fixed with 5026298796a68dd3eb1d6b7731e25475bf4e67fe.

PR #155, opened 2016-06-06
Add support for locate module

chiro79:
Adds the full commands for locate module, which includes:
- locate
- updated
- stats
- version

Also this commit adds a test class and an example executable class using "locate"
command.

It was tested with version 2015.8.7 which for me the locate module was not working
and I had to fix the python file ﻿"salt/modules/locate.py":
```
﻿111,113c111,112
<         if option in kwargs:
<             if bool(kwargs[option]) is True:
<                 options += toggles[option]
---
>         if bool(kwargs[option]) is True:
>             options += toggles[option]
```

The style of the code generated tries to be coherent with the existent one.


src/main/java/com/suse/salt/netapi/calls/modules/Locate.java
```diff
@@ -0,0 +1,182 @@
...
+    private static final LocalCall<Stats> STATS = new LocalCall<>("locate.stats",
+            Optional.empty(), Optional.empty(),
+            new TypeToken<Stats>() {
+            });
+
+    /**
+     * Locate module result object
+     */
+    public static class Stats {
+        private Long files;
```
lucidd: Is there a reason you are using boxed types everywhere?

chiro79: Hello,
I usually use primitive types except in some special scenarios. With this API I just wanted to be consistent with the rest of the code. When I was developing I looked at "com.suse.salt.netapi.calls.modules.File" class and it has boxed types for the result class. But checking the rest of the code again after you comment, it is true that in other classes the primitive types are used.

So, do you want me to change it to primitive types?

lucidd:
We generally use primitive values whenever possible. I think the `Result` class may have slipped through code review. It would be good if you could change to primitive values where possible.

src/main/java/com/suse/salt/netapi/calls/modules/Locate.java
lucidd: We generally use primitive values whenever possible. I think the `Result` class may have slipped through code review. It would be good if you could change to primitive values where possible.

canbulutb: The reason for the boxed `Boolean` type in `Result` is to support tri-state booleans exposed in netapi, which is indeed the case in some File module methods (`True`/`False`/`None`).

However, it turns out `None` is returned only when the API is called with the `test` argument, which we don't actually expose. So, I'll post a PR to fix the `Result` class, too.

chiro79:
By the way, the fix for the locate module is wrong in the commit, I did a one little more change, the diff between the fix and the original is:
```
110,113c110,112
<     for option in toggles:
<         if option in kwargs:
<             if bool(kwargs[option]) is True:
<                 options += toggles[option]
---
>     for option in kwargs:
>         if bool(kwargs[option]) is True:
>             options += toggles[option]
```

src/main/java/com/suse/salt/netapi/calls/modules/Locate.java
lucidd: i'm not really a fan of `null` as a valid state. We usually use `Optional` to avoid the use of `null` at least in the user facing API.

src/main/java/com/suse/salt/netapi/calls/modules/Locate.java
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+
+    /**
+     * All possible options for "locate" method.
+     */
+    public static class LocateOpts {
+        private Map<String, Boolean> opts = new HashMap<>();
+
+        public void setBasename(Boolean basename) {
```
lucidd: here we should also use primitives.

chiro79: OK, I agree.
I thought that because the value is going to be boxed anyway when it is stored in the Map, it would be acceptable to request a Boolean instead of boolean. But checking other parts of the code in where a parameter exposed by the API is stored in a map, I have seen that a primitive type is requested. For example, here: com/suse/salt/netapi/calls/modules/File.java:75

I'll change it tonight.

lucidd: You are right that it will be boxed anyway but accepting a boxed value instead of a primitive opens up the possibility to pass `null` which is not valid in this case. So it just to remove ways to misuse the API.

renner:
I am closing this one now in favor of #157, everything here seems to be contained there as well IIUC.

PR #157, opened 2016-06-08
Null collections fix

chiro79:
This branch is based on the "module-locate" one. Please check my comments on the commits 6f7ade3 and a1ebb77.


src/main/java/com/suse/salt/netapi/parser/CollectionTypeAdapterFactory.java
```diff
@@ -0,0 +1,82 @@
...
+        private final ObjectConstructor<? extends Collection<E>> constructor;
+
+        public Adapter(TypeAdapter<E> elementAdapter, ObjectConstructor<? extends
+                Collection<E>> constructor) {
+            this.constructor = constructor;
+            this.elementTypeAdapter = elementAdapter;
+        }
+
+        public Collection<E> read(JsonReader in) throws IOException {
+            Collection<E> collection = constructor.construct();
```
lucidd: we should also explicitly check for `null` here like we do in other adapters.

lucidd:
LGTM :+1: thanks for the quick adjustment

chiro79:
Please, take into account that I did it in a limited time just to show you a problem and a first try to fix it. This class is just a copy of the Gson one with a Json null check deleted:
```
public Collection<E> read(JsonReader in) throws IOException {
      if (in.peek() == JsonToken.NULL) {
        in.nextNull();
        return null;
}
```

Maybe there is a better and wide valid solution like to having in one point a check for JSON elements with null value and in there generate the GenericSaltError.

How do you see it?

src/main/java/com/suse/salt/netapi/parser/CollectionTypeAdapterFactory.java
chiro79: I have checked the code, and in some Adapters there is no check, for example in the "Adapters" or "DateAdapter" classes all reads are no checked.
But, checking other Adapters you are right and null is checked. I'll change it tonight adding:
```
        if (jsonReader.peek() == JsonToken.NULL) {
            throw new JsonParseException("null is not a valid value for a Collection");
        }
```
Is that ok?

lucidd:
there may be a general solution by having a type adapter factory that hijacks all types and wraps the original one with the null check. But i don't think it would work because you have to lookup the original adapter which probably would trigger the same type adapter we are already in. 
We could also try to wrap the adapters / factories before we register them.

lucidd:
We merge this now since we currently preparing for a new release and your changes look good. If there are any improvements you can think of feel free to open another PR. In any case thanks for the great work :smile: :+1:

PR #160, opened 2016-06-13
add jobs.list_job(s) and saltutil.running

lucidd:
this add the functions `jobs.list_job(s)` and `saltutil.running`


renner:
Merged.

PR #161, opened 2016-06-22
Synchronous module calls via salt-ssh

renner:
This patch adds support for calling `LocalCall` instances synchronously via `salt-ssh`. The results of calls to `state.apply` were further refined to be parsed into the new type `State.ApplyResult` that contains the changes unparsed as `JSONElement`.


renner:
@lucidd: Could you please have another look at the patch before I merge it?

PR #167, opened 2016-07-15
Change type of `registerrelease`

mbologna:
Using salt 2015.8.7 (Beryllium), field `registerrelease` is marked as boolean.

Example:
2016-07-15 14:03:04,511 [RHN Message Dispatcher] WARN  com.suse.manager.webui.services.impl.SaltAPIService - GenericSaltError([{"eol_t":1730332800,"productline":"sles","name":"SLES","installed":true,"vendor":"SUSE","summary":"SUSE Linux Enterprise Server 12 SP1","repo":"@System","registerrelease":false,"epoch":"0","version":"12.1","release":"0","isbase":true,"flavor":"DVD","eol":"2024-10-31T00:00:00+00","arch":"x86_64","shortname":"SLES12-SP1","description":"SUSE Linux Enterprise offers a comprehensive suite of products built on a single code base. The platform addresses business needs from the smallest thin-client devices to the world's most powerful high-performance computing and mainframe servers. SUSE Linux Enterprise offers common management tools and technology certifications across the platform, and each product is enterprise-class."}], java.lang.IllegalStateException: Expected STRING but was BOOLEAN)

### DIFF
--- a/src/main/java/com/suse/salt/netapi/calls/modules/Zypper.java
+++ b/src/main/java/com/suse/salt/netapi/calls/modules/Zypper.java
@@ -29,7 +29,7 @@ public class Zypper {
         private final boolean installed;
         private final boolean isbase;
         private final String productline;
-        private final String registerrelease;
+        private final boolean registerrelease;
         private final String release;
         private final String repo;
         private final String shortname;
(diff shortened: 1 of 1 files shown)


moio:
@lucidd are we aware of any Salt version where this patch would break, ie. where `registerrelease` is really a `String`?

renner:
@moio: I can perfectly reproduce the issue running `2015.8.7 (Beryllium)` minion and master. I think this code was working with an older version of `salt` though, but I don't know which one was that.

mbologna:
Closing this PR because the problem has been fixed on Salt, so we should use the current behavior: https://bugzilla.suse.com/show_bug.cgi?id=989193

PR #170, opened 2016-07-22
Major API cleanup

renner:
This patch is going to bring a major API cleanup: Some of the [Salt API endpoints](https://docs.saltstack.com/en/latest/ref/netapi/all/salt.netapi.rest_cherrypy.html) are only shortcuts to certain execution modules, e.g. the `/minions` or `/jobs` endpoints. We would like to unify and simplify the API of this library by removing support for targeting all the Salt API endpoints. Instead all calls will use either `/run` or `/` assuming that all functionality can be implemented by using those endpoints with one of either `LocalCall`, `WheelCall` or `RunnerCall`.


renner:
Finally merged, after 2 years!

PR #171, opened 2016-07-25
Added local_batch support with tests.

### LINKED ISSUE #84: Support additional client types
Salt's api supports additional methods of interaction. Specially, it would be helpful to support local_batch and runner in addition to the existing local client type. 

local_batch would allow for using the batch interface, while runner would allow for kicking off orchestrate jobs. 

http://docs.saltstack.com/en/latest/topics/netapi/index.html#client-interfaces

briancecker:
Fixes #84 

Refactored some code in LocalCall because callSync and callSyncBatch
used nearly the same logic. They now both use a helper function called
callSyncBatchable.

Added a test for callSyncBatch that checks the request body. Also added
a test for callSync that checks the request body to ensure it is not
accidentally using batching when calling callSyncBatchable.

### DIFF
--- a/src/main/java/com/suse/salt/netapi/calls/LocalCall.java
+++ b/src/main/java/com/suse/salt/netapi/calls/LocalCall.java
@@ -4,6 +4,7 @@ import static com.suse.salt.netapi.utils.ClientUtils.parameterizedType;
 
 import com.suse.salt.netapi.AuthModule;
 import com.suse.salt.netapi.client.SaltClient;
+import com.suse.salt.netapi.datatypes.Batch;
 import com.suse.salt.netapi.datatypes.target.Target;
 import com.suse.salt.netapi.exception.SaltException;
 import com.suse.salt.netapi.results.Result;
@@ -143,8 +144,27 @@ public class LocalCall<R> implements Call<R> {
      */
     public Map<String, Result<R>> callSync(final SaltClient client, Target<?> target)
             throws SaltException {
-        return callSyncBatchable(client, target, Optional.empty(),
+        return callSyncHelper(client, target, Optional.empty(),
                 Optional.empty(), Optional.empty(), Optional.empty()).get(0);
+    }
+
+    /**
+     * Calls a execution module function on the given target with batching and
+     * synchronously waits for the result. Authentication is done with the token
+     * therefore you have to login prior to using this function.
+     *
+     * @param client SaltClient instance
+     * @param target the target for the function
+     * @param batch  the batch specification
+     * @return A list of maps with each list representing each batch, and maps containing
+     * the results with the minion names as keys.
+     * @throws SaltException if anything goes wrong
+     */
+    public List<Map<String, Result<R>>> callSync(final SaltClient client, Target<?> target,
+            Batch batch)
+            throws SaltException {
+        return callSyncHelper(client, target, Optional.empty(),
+                Optional.empty(), Optional.empty(), Optional.of(batch));
     }
 
     /**
@@ -164,27 +184,8 @@ public class LocalCall<R> implements Call<R> {
             final SaltClient client, Target<?> target,
             String username, String password, AuthModule authModule)
             throws SaltException {
-        return callSyncBatchable(client, target, Optional.of(username),
+        return callSyncHelper(client, target, Optional.of(username),
                 Optional.of(password), Optional.of(authModule), Optional.empty()).get(0);
-    }
-
-    /**
-     * Calls a execution module function on the given target with batching and
-     * synchronously waits for the result. Authentication is done with the token
-     * therefore you have to login prior to using this function.
-     *
-     * @param client SaltClient instance
-     * @param target the target for the function
-     * @param batch  the batch parameter, e.g. 10, 10%, etc.
-     * @return A list of maps with each list representing each batch, and maps containing
-     * the results with the minion names as keys.
-     * @throws SaltException if anything goes wrong
-     */
-    public List<Map<String, Result<R>>> callSyncBatch(
-            final SaltClient client, Target<?> target, String batch)
-            throws SaltException {
-        return callSyncBatchable(client, target, Optional.empty(),
-                Optional.empty(), Optional.empty(), Optional.of(batch));
     }
 
     /**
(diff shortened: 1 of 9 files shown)


src/main/java/com/suse/salt/netapi/calls/LocalCall.java
```diff
@@ -178,13 +164,85 @@ public LocalCall(String functionName, Optional<List<?>> arg,
...
+     * therefore you have to login prior to using this function.
+     *
+     * @param client SaltClient instance
+     * @param target the target for the function
+     * @param batch  the batch parameter, e.g. 10, 10%, etc.
+     * @return A list of maps with each list representing each batch, and maps containing
+     * the results with the minion names as keys.
+     * @throws SaltException if anything goes wrong
+     */
+    public List<Map<String, Result<R>>> callSyncBatch(
```
lucidd: I think its nicer if this method is just an overloaded version of `callSync` with the addition of the `batch` parameter. No need to put it in the name.

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
```diff
@@ -178,13 +164,85 @@ public LocalCall(String functionName, Optional<List<?>> arg,
...
+            final SaltClient client, Target<?> target, String batch)
```
lucidd: Even if salt represents batch as a string i think it a little bit too broad and error prone. It would be nice to have it be a real type + some helpers to construct them like `Batch.percent(20)` or `Batch.amount(200)`.

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
briancecker: Hi, thanks for the feedback. I saw that you had mentioned simply overloading the `callSync` methods in a previous comment, but I noticed that the other values for the `Client` enum seemed to have separate methods as well, so I followed the pattern, e.g. `Client.LOCAL` has `callSync`, `Client.LOCAL_ASYNC` has `callAsync`, `Client.SSH` has `callSyncSSH`.

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
briancecker: I was considering that, and I'll push a new commit with that in mind.

lucidd:
Thanks for this PR i'm happy someone is finally tackling this feature :smile:

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
lucidd: Good point. Async is a little bit different since it does not have an additional parameter to differentiate itself the only difference is the return type and its not possible to overload on different return type alone. The SSH one should probably be renamed too (not in this PR tho).

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
briancecker: @lucidd About the Batch parameter, technically Salt allows batches of `"0%"` and `"0"`, but they just cause the system to hang in a loop I believe, so I'm just going to restrict those to ` > 0`.

lucidd:
@briancecker looks better now :+1:

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
```diff
@@ -1,23 +1,24 @@
...
-
 import java.lang.reflect.Type;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import static com.suse.salt.netapi.utils.ClientUtils.parameterizedType;
+
+
```
renner: Please don't (let your IDE) reorder imports randomly, same above with `TypeToken`. Your patch should contain only changes that are actually relevant.

src/test/java/com/suse/salt/netapi/calls/LocalCallTest.java
```diff
@@ -1,35 +1,35 @@
...
+import static com.github.tomakehurst.wiremock.client.WireMock.equalToJson;
+import static com.github.tomakehurst.wiremock.client.WireMock.postRequestedFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlMatching;
+import static com.github.tomakehurst.wiremock.client.WireMock.verify;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
```
renner: Same as before, please don't reorganize imports if not relevant for the patch.

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
```diff
@@ -143,33 +144,38 @@ public LocalCall(String functionName, Optional<List<?>> arg,
...
      * no session token is created.
      *
-     * @param client SaltClient instance
-     * @param target the target for the function
-     * @param username username for authentication
-     * @param password password for authentication
+     * @param client     SaltClient instance
+     * @param target     the target for the function
+     * @param username   username for authentication
+     * @param password   password for authentication
```
renner: Please don't (let your IDE) introduce additional whitespace that is not relevant for the actual patch.

briancecker: @renner Thanks, I've changed my IDE settings so it doesn't "fix" things for me. 

What is the preferred work flow in regards to modifying commits and force pushing? I've squashed the commits into a single one and reverted the whitespace changes. Is it preferred to force push to this branch/PR, or is it preferred to simply open a new PR and link to it from here?

renner: @briancecker: Do not open another PR to improve this patch. It is OK to force push this PR if you like, but from my perspective there is no need to end up with just a single commit. I personally prefer to just add more commits fixing up things, but do as you like.

briancecker:
@renner I rebased and removed all the unnecessary whitespace changes. The changes should be cleaner now.

renner:
Thanks, @briancecker! It looks good to me now, merged it.

PR #174, opened 2016-09-13
Add support for `raw_shell` commands via salt-ssh

mbologna:
Current implementation for `salt-ssh` raw command execution lacks the parameter for launching the command.
This PR implements "raw shell" command execution for `salt-ssh` in another method which will be used for raw command only.

Reference: https://docs.saltstack.com/en/latest/topics/ssh/#raw-shell-calls

### DIFF
--- a/src/main/java/com/suse/salt/netapi/client/SaltClient.java
+++ b/src/main/java/com/suse/salt/netapi/client/SaltClient.java
@@ -411,14 +411,14 @@ public class SaltClient {
         return result.getResult().get(0);
     }
 
-    public <T> Map<String, Result<SSHRawResult>> runRawSSHCommand(final Target<T> target,
-            final String function)
+    public <T> Map<String, Result<SSHRawResult>> runRawSSHCommand(final String command,
+            final Target<T> target)
         throws SaltException {
         Map<String, Object> props = new HashMap<>();
         props.put("client", Client.SSH.getValue());
         props.put("tgt", target.getTarget());
         props.put("expr_form", target.getType());
-        props.put("fun", function);
+        props.put("fun", command);
         props.put("raw_shell", true);
 
         List<Map<String, Object>> list = Collections.singletonList(props);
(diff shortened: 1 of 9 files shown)


src/main/java/com/suse/salt/netapi/results/SSHRawResult.java
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Wrapper class for salt-ssh results.
+ *
+ */
+public class SSHRawResult {
+
+    private int retcode;
+    private Optional<String> stderr = Optional.empty();
+    private Optional<String> stdout = Optional.empty();
```
lucidd: i'm not sure but wasn't the raw ssh the one where stdout/stderr is always present regardless of success or failure?

src/main/java/com/suse/salt/netapi/client/SaltClient.java
```diff
@@ -409,6 +416,27 @@ public ResultInfoSet getJobResult(final String job) throws SaltException {
         return result.getResult().get(0);
     }
 
+    public <T> Map<String, Result<SSHRawResult>> runRawSSHCommand(final Target<T> target,
```
lucidd: we still need the ssh config here right? also i think it would be better to have the command first and then the target.

src/test/resources/ssh_raw_run_response.json
```diff
@@ -0,0 +1,11 @@
...
+  "return": [
+    {
+      "sumarm30": {
+        "retcode": 0,
+        "stderr": "Could not create directory '/srv/salt/.ssh'.\r\nFailed to add the host to the list of known hosts (/srv/salt/.ssh/known_hosts).\r\n",
+        "stdout": "Password: \n 17:42pm  up  19:16,  0 users,  load average: 0.00, 0.01, 0.05\n"
+      }
+    }
+  ]
+}
```
lucidd: missing new line

PR #175, opened 2016-11-04
Prevent wrong target types from being passed to salt ssh

cbbayburt:
Salt ssh has a limited set of target types it can handle (compared to normal salt) see https://docs.saltstack.com/en/latest/topics/ssh/#targeting-with-salt-ssh.

Due to this its currently possible to call salt ssh in our netapi library with parameters that will not work.

The plan is to only accept target types that salt ssh can handle.
One way to achieve this would be an additional marker interface that is only implemented by targets that can be used with salt ssh.

### DIFF
--- a/src/main/java/com/suse/salt/netapi/datatypes/target/Glob.java
+++ b/src/main/java/com/suse/salt/netapi/datatypes/target/Glob.java
@@ -3,7 +3,7 @@ package com.suse.salt.netapi.datatypes.target;
 /**
  * Target for specifying minions by glob pattern.
  */
-public class Glob implements Target<String> {
+public class Glob implements Target<String>, SSHTarget<String> {
 
     public static final Glob ALL = new Glob("*");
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/suse/salt/netapi/datatypes/target/Glob.java
```diff
@@ -3,7 +3,7 @@
 /**
  * Target for specifying minions by glob pattern.
  */
-public class Glob implements Target<String> {
+public class Glob implements Target<String>, SshTarget<String> {
```
cbbayburt: `SshTarget` already extends `Target`, but `Target` is also included here just for readability. Otherwise it may look like Glob is *exclusively* used for salt-ssh.

hustodemon:
LGTM

renner:
I would request one very nitpicky change: Can we make `SSH` in the interface name all uppercase, like `SSHTarget`? The reason is that we already have other classes with `SSH` in the name so I would like to be consistent there. Also it looks better in my opionion. Thanks!

PR #178, opened 2016-11-21
Implementation of callAsync that resolves results via eventstream

lucidd:
This implements a callAsync method that sets of a normal salt async call and then listens for the results on the eventstream. Since there is a gab between the call and installing the listeners (which require the jid returned by the async call) we do an additional job lookup after all listeners are installed. The listeners then look for either the job return event or the job lookup event.

@mateiw @mbologna


src/main/java/com/suse/salt/netapi/calls/LocalCall.java
```diff
@@ -101,6 +145,76 @@ public LocalCall(String functionName, Optional<List<?>> arg,
...
+                                            new GenericError(
+                                                    "EventStream closed with reason "
+                                                            + closeReason)
+                                    ));
+                                }
+                            };
+                            events.addEventListener(listener);
+                            cancel.whenComplete((v, e) -> {
+                                if (v != null) {
+                                    future.complete(Result.error(v));
```
mateiw: If `cancel` completes exceptionally `v` will be null and `future` won't complete at all.

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
```diff
@@ -101,6 +145,76 @@ public LocalCall(String functionName, Optional<List<?>> arg,
...
+     * as they come in via the event stream.
+     *
+     * @param client SaltClient instance
+     * @param target the target for the function
+     * @param events the event stream to use
+     * @param cancel future to cancel the action
+     * @return a map from minion id to future of the result.
+     * @throws SaltException if anything goes wrong
+     */
+    public Map<String, CompletionStage<Result<R>>> callAsync(
```
mateiw: I think it would be useful to have an additional method, similar to this one, that takes the username and password.

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
lucidd: i was wondering about that. But you are right if we don't cancel it in this case it may never be canceled.

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
lucidd: we could do that but the only way to get an event stream is by token authentication so if you pass it an event stream you should be already authenticated. We could do the username password thing instead of the event stream and then creating the event stream ourselfs. What do you think?

renner: @lucidd: that sounds better, but it would mean that every call to this `callAsync()` method would open a separate websocket connection, right?

lucidd: yes it would mean that. I think we will just provide several overloads of this method to cover different use cases.

mbologna: +1 for @mateiw proposal

lucidd:
@mateiw the last commit changes the implementation to only use one listener and also provides multiple overloads including one for adhoc authentication.

MalloZup:
cucumber test is there: https://github.com/SUSE/spacewalk-testsuite-base/pull/126, i will merge it once the feature got merged

mateiw:
@lucidd your last commit LGTM 👍

renner:
@lucidd: Let me know when you are ready to merge this, from my perspective it is good to be merged.

lucidd:
@renner i talked to @hustodemon and it seems that we will need to add another error type for salt ssh.
We can merge this if you want its just abusing the JsonParsingError a but until we have a better error type for that.

PR #180, opened 2016-12-20
status.uptime was not working

smatei:
Hi,

Status.uptime was not working. https://github.com/SUSE/salt-netapi-client/issues/179

I removed the human_readable argument, as it was reported invalid. Changed the return type from Float to String.

Stefan

### DIFF
--- a/src/main/java/com/suse/salt/netapi/calls/modules/Status.java
+++ b/src/main/java/com/suse/salt/netapi/calls/modules/Status.java
@@ -6,7 +6,6 @@ import com.google.gson.reflect.TypeToken;
 
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


renner:
Hi @smatei, thanks for your pull request! What exact version of Salt did you test this with? For versions of Salt >= 2015.8.9 a [dict should be returned rather than just a String](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.status.html#salt.modules.status.uptime). Shouldn't we actually support this newer format?

smatei:
Hi @renner. You are right, the version was old, 2015.5.10. I have installed 2016.11.1 and changed response type to Map<String, Object>, now I get 

{ns3024987.ip-149-202-197.eu=Result({users=3.0, seconds=1.0082822E7, since_t=1.472173602E9, days=116.0, since_iso=2016-08-26T01:06:42.256637, time=16:47})}

Is this OK?

renner:
Okay, I am merging it like this. Of course it will work only with Salt versions >= 2015.8.9, but that should be fine. A unit test would be nice to have, but we can add it later in a separate pull request. Thanks for your patch, @smatei!

smatei:
@renner 
I see that there is no unit test for com.suse.salt.netapi.calls.modules.Status. I will add one starting with status.uptime. And then add tests for other status methods.

You can close the issue https://github.com/SUSE/salt-netapi-client/issues/179.

PR #181, opened 2016-12-24
Added test for status.uptime command

smatei:
Hi @renner 

As I promised here, https://github.com/SUSE/salt-netapi-client/pull/180, I have added a test for status.uptime command. I will add test for other status commands too.


renner:
Thanks for your contribution @smatei, I merged it.

PR #186, opened 2017-01-25
Added salt.modules.minion module.

smatei:
Added salt.modules.minion. It is useful for listing the master's minions when client.getMinions() throws "SaltUserUnauthorizedException: Salt user does not have sufficient permissions".


renner:
Hey @smatei, sorry for the late reply and thank you for this contribution! I merged to have this included with the `0.10.0` release that we are currently preparing.

Issue #163, opened 2016-07-02
WARNING: Invalid cookie header


cinterloper:
Is it normal to receive log messages like this?

```
Jul 01, 2016 10:32:55 PM org.apache.http.client.protocol.ResponseProcessCookies processCookies
WARNING: Invalid cookie header: "Set-Cookie: session_id=_MASKED_INFO_; expires=Sat, 02 Jul 2016 15:32:55 GMT; Path=/". Invalid 'expires' attribute: Sat, 02 Jul 2016 15:32:55 GMT
salt auth token: _MASKED_INFO_ Until: Sat Jul 02 10:32:55 MST 2016 Perms : [.*, @runner, @wheel, @jobs]
Jul 01, 2016 10:32:55 PM org.apache.http.client.protocol.ResponseProcessCookies processCookies
WARNING: Invalid cookie header: "Set-Cookie: session_id=_MASKED_INFO_; expires=Sat, 02 Jul 2016 15:32:55 GMT; Path=/". Invalid 'expires' attribute: Sat, 02 Jul 2016 15:32:55 GMT
```

renner:
If haven't seen those warnings yet, what exactly would I do to reproduce this?

cinterloper:
This occurs for me when i execute this piece of code
`https://github.com/cinterloper/vertx-salt/blob/master/src/main/resources/saltConnection.groovy`

thats an example that should connect to salt-api on localhost:8000 plaintext. 

the salt master i have configured with the following stanza:
```
rest_cherrypy:
  port: 8000
  debug: True
  disable_ssl: True
  expire_responses: False
```

to build that project (vertx-salt)
./gradlew clean shadoJar
to run 
 java -jar build/libs/vertx-salt-3.3.0-5-fat.jar

cinterloper:
fwiw, this is the full salt-master config
```
rest_cherrypy:
  # can be any port
  port: 8000
  debug: True
  disable_ssl: True
  #ssl_crt: /etc/pki/tls/certs/localhost.crt
  #ssl_key: /etc/pki/tls/certs/localhost.key
  expire_responses: False
  
file_recv: True

ext_pillar:
  - file_tree:
      root_dir: /srv/pillar/tree
      follow_dir_links: False
      raw_data: False

file_roots:
  base:
    - /srv/salt

client_acl:
  user:
    - test.ping

external_auth:
  pam:
    user:
      - .*
      - '@runner'
      - '@wheel'
      - '@jobs'

[...]

PR #189, opened 2017-02-15
Fixed Cookie issue

adohren:
Hello,

I checked the cookie issue. The httpclient uses default policy, if you don't define any other one. So I changed it to use standard policy that can handle the date format.

What do you think about it?

### DIFF
--- a/src/main/java/com/suse/salt/netapi/client/impl/HttpClientConnection.java
+++ b/src/main/java/com/suse/salt/netapi/client/impl/HttpClientConnection.java
@@ -1,10 +1,8 @@
 package com.suse.salt.netapi.client.impl;
 
-import com.suse.salt.netapi.client.Connection;
-import com.suse.salt.netapi.config.ClientConfig;
-import com.suse.salt.netapi.exception.SaltException;
-import com.suse.salt.netapi.exception.SaltUserUnauthorizedException;
-import com.suse.salt.netapi.parser.JsonParser;
+import java.io.IOException;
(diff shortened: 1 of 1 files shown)


renner:
Hey @adohren, thanks for your patch! Assuming that you are referring to #163: are you possibly aware of a more simple reproducer for the issue so I can test your fix or did you run @cinterloper's groovy code?

src/main/java/com/suse/salt/netapi/client/impl/HttpClientConnection.java
```diff
@@ -23,9 +22,11 @@
...
 import org.apache.http.impl.client.HttpClients;
 
-import java.io.IOException;
-import java.io.UnsupportedEncodingException;
-import java.net.URI;
+import com.suse.salt.netapi.client.Connection;
+import com.suse.salt.netapi.config.ClientConfig;
+import com.suse.salt.netapi.exception.SaltException;
+import com.suse.salt.netapi.exception.SaltUserUnauthorizedException;
+import com.suse.salt.netapi.parser.JsonParser;
```
renner: Please don't let your IDE reorganize all imports just for fun, in this case there should be only one additional import (`CookieSpecs`).

renner:
@adohren: We are currently about to prepare a new release and we might want to include this patch. Would you mind to implement the changes as requested by me?

renner:
@adohren Nevermind, cherry-picked and as 3bd1213f1502b6d797a52647676b38cb4b278547.

Issue #149, opened 2016-04-23
Add module name information to Call classes.


lucidd:
This will allow us to use the call objects to generate input for functions like `state.high` and `state.low`.

admd:
I would appreciate if you can provide a bit more information regarding this requirement.

lucidd:
The general idea of this was to have the `*Call` classes know about the module name of the function they are calling. There are multiple places where is would be helpful to know the module name. For example the `state.high` and `state.low` could potentially be used to to bundle multiple `LocalCall`s into one input for `state.high/low` via `module.run`. But to generate this input we need to know the modue name of each `LocalCall` we want to include.

admd:
Thanks @lucidd for the quick reply. While looking into this issue, i see that there are some duplication in *Call classes. Would it make sense to move the common code to an abstract class and extend our *Call classes from that?

admd:
Ok back to the original issue.  Two scenarios.
1st, keep  the old constructor(for backward compatibility) where module name is appended to function like {modulename}.{function} and then get the module name from this parameter by some parsing rule.

2nd, add another constructor which accepts module name as a separate parameter to function.

Furthermore, function parameter can also be changed to an array of  strings or varargs like {high, low} if want to pass more than one function. Also, instead of functions as Strings we can introduce Enums with all the functions of a module.

Pleas bear with me if am writing some not so intelligent stuff, i haven't get much chance to contribute in community before so ignore it if anything is rude or so. It's going to be my 1st PR in community project so i am way more conscious in my approach.

lucidd:
@admd if there is common code that can be put in a super class that would be very welcome. :+1:

renner:
After the merge of #197 this issue is partly *fixed*: we now have the module information available, but we still would want to have a getter for only the function name. Note that the current `getFunction()` returns the whole string as composed from `module.function`.

admd:
I will take care of this on weekend .

PR #193, opened 2017-03-03
Refactored some code and added module name information to Call classes

admd:
#149

### DIFF
--- a/src/main/java/com/suse/salt/netapi/calls/LocalCall.java
+++ b/src/main/java/com/suse/salt/netapi/calls/LocalCall.java
@@ -51,11 +51,10 @@ public class LocalCall<R> extends AbstractCall<R> {
         this.metadata = metadata;
     }
 
-
     public LocalCall(String moduleName, String functionName, Optional<List<?>> arg,
                      Optional<Map<String, ?>> kwargs, TypeToken<R> returnType,
                      Optional<?> metadata) {
-        super(moduleName,functionName, kwargs, returnType);
+        super(moduleName, functionName, kwargs, returnType);
         this.arg = arg;
         this.metadata = metadata;
     }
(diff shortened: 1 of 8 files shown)


admd:
I am not sure why this one test is failing.

admd:
I have get the latest master version without my changes and test is still failing there in my local environment. 
What I can't understand how come Travis was building fine before my changes.

admd:
@lucidd can you please review this PR and let me know if i have to change anything.

admd:
@lucidd Thanks for the review, i have corrected the changes. Sorry for the late reply. I was out for a short vacation.

renner:
@admd, @lucidd: I manually merged and solved some conflicts caused by the merge of #195. You might want to have a look at the current master to check if I missed something.

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
```diff
@@ -38,50 +38,49 @@
...
     /**
      * {@inheritDoc}
      */
     @Override
     public Map<String, Object> getPayload() {
-        HashMap<String, Object> payload = new HashMap<>();
-        payload.put("fun", functionName);
+        Map<String, Object> payload = super.getPayload();
         arg.ifPresent(arg -> payload.put("arg", arg));
-        kwarg.ifPresent(kwarg -> payload.put("kwarg", kwarg));
```
renner: @admd and @lucidd: After we merged this patch for releasing it with 0.11.0 we found that this specific change broke things elsewhere as for LocalCall it has always been `kwarg` (without "s")! We need to verify again here if it is actually correct to turn this into `kwargs` now for all call types, why it has been `kwarg` so far and if we maybe need to keep it.

renner:
@admd: After merging this patch we had to revert it again as we were unsure about the change of `kwarg` into `kwargs` for `LocalCall`, see my inline comment. We released `0.11.1` without it, sorry! Unfortunately I cannot reopen this PR as the commits were actually merged. Could you please make a new PR against the current master so we can go on from there? Sorry again for this inconvenience.

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
lucidd: I remember when implementing this that `kwarg` and `kwargs` are not compatible. I'm not exactly sure why salt has 2 different names here. In the worst case we just have to move the kwargs to the specific implementation or have a abstract getter for the name which can be overridden.

admd: Thanks for the feedback and sorry for the inconvenience caused by my change. Somehow tests gave me wrong impression that my code didn't break anything, I should have been more careful.
 
I will make changes suggested above through a new PR.

renner: Ping @admd, would you mind to create that new PR so that we can go on from there? Thank you!

admd: Hi @renner ,  I have created another PR and excluded those changes which caused the issue last time. Code can still be imroved but because of lack of time it's difficult right now for me. I will create another PR for further refactoring . 

For Issue #149 , this PR should be enough.

PR #195, opened 2017-03-15
Allowing custom timeouts for Manage.up(), Test.ping() and on a generic LocalCall

meaksh:
This PR allows to set custom `timeout` and `timeout_job_timeout` salt attributes when calling `Manage.up()`, `Test.ping()` or a generic `LocalCall`.

This is very useful when some minions are unreachable and we want to get a presence report in a short and fixed amount of time.

Salt upstream PR: https://github.com/saltstack/salt/pull/40018


meaksh:
@lucidd @mbologna - This is ready to get some feedback. Please, take a look if you have some time.

mbologna:
Looks good to me, only thing missing are junit tests (which can be trivial).

moio:
Tests added, this also LGTM. Assigning to @lucidd to get his ultimate approval.

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
```diff
@@ -83,6 +102,9 @@ public LocalCall(String functionName, Optional<List<?>> arg,
         arg.ifPresent(arg -> payload.put("arg", arg));
         kwarg.ifPresent(kwarg -> payload.put("kwarg", kwarg));
         metadata.ifPresent(m -> payload.put("metadata", m));
+        timeout.ifPresent(timeout -> payload.put("timeout", timeout));
```
lucidd: is `timeout` and `gather_job_timeout` supported on every `local_*` client?

src/main/java/com/suse/salt/netapi/calls/modules/Test.java
```diff
@@ -138,6 +138,12 @@ public VersionInformation(Map<String, String> salt,
         return PING;
     }
 
+    public static LocalCall<Boolean> ping(Optional<Integer> timeout,
```
lucidd: i don't think we should make `ping` a special case here if those parameters are supported on every function. We should rather do something along the lines of `withMetadata`.

src/main/java/com/suse/salt/netapi/calls/runner/Manage.java
```diff
@@ -27,6 +27,15 @@ private Manage() { }
                 new TypeToken<List<String>>(){});
     }
 
+    public static RunnerCall<List<String>> up(Optional<Integer> timeout,
+            Optional<Integer> gatherJobTimeout) {
```
lucidd: is there a semantical difference between the parameters to `up` and the now generally introduced ones? what happens if you specify both sets of parameters? which one will win?

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
meaksh: Well, these two parameters are now handled inside `salt.client.LocalClient` class of Salt, which is the class that handles all `local_*` clients.

The `local`, `local_async` clients will work as expected, but looking now at `local_batch` in the Salt side, it seems that kwargs won't be propagated to the Batch object so I'll expect that timeouts parameters won't work on this case. I'll fix this on Salt ASAP. Thanks! :+1:

moio: I would say that the Batch case should be fixed in Salt, as ATM it is a Salt bug. That should not block this PR anyway, as the library is doing the  right thing FMPOV.

Is that OK with you @lucidd?

src/main/java/com/suse/salt/netapi/calls/modules/Test.java
moio: For the record, we discussed this with @meaksh and he will adjust this PR as per your suggestion. Thanks

src/main/java/com/suse/salt/netapi/calls/runner/Manage.java
moio: No, their semantics are the same - but in this case we cannot introduce this in a general way for all RunnerCalls, simply because those two parameters have been added "ad hoc" for [manage.up](https://github.com/saltstack/salt/pull/40018/files#diff-7ab664360b664be453b91fe52517b5a3R137) and [manage.status](https://github.com/saltstack/salt/pull/40018/files#diff-7ab664360b664be453b91fe52517b5a3R33).

@lucidd: is it OK to add them as Optionals or would method overloading make more sense here to be backwards compatible?

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
lucidd: if it will be fixed in salt then its ok. Only in case it will or cant be fixed in salt we should make sure in the library that you cant end up configuring something that will never work the way you expect.

src/main/java/com/suse/salt/netapi/calls/runner/Manage.java
lucidd: Using `Optional` in most module functions is more because of lazyness. Overloading is always preferred but gets quickly out of hand in case of many optional parameters. We don't promise backwards compatibility before 1.0 but if its easy to provide in this case we should go for it.

meaksh:
@lucidd - Thank you for your feedback in this PR. I've added few more commits to address the discussion here:

- Added `withTimeouts` and `withoutTimeouts` in a similar way as `withMetadata`.
- Fixed `withMetadata` to preserve timeouts values if they exists.
- Got rid of `Test.ping(timeout, gatherJobTimeout)`.
- Fixed unit test.

Please, take a look when you have time.

Still pending to fix the Batch issue. I would expect a PR to upstream Salt in the following days since I pointed this out in a Trello card.

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
meaksh: @lucidd - This PR fixes `local_batch` client: https://github.com/saltstack/salt/pull/40300

PR #197, opened 2017-04-13
Refactored some code and added module name information to Call classes

admd:
LocalCall uses 'kwarg' parameter while RunnerCall and WheelCall use 'kwargs'(with s ), during last PR i renamed 'kwarg' parameter to 'kwargs' parameter and extracted that parameter in a parent class which was wrong and caused some issues. I have reverted those changes back. Code can definitely be improved further but my lack of Salt knowledge and right now being on holidays is kind of hindrance for the time being.

Related discussion can be seen in  #193 . Main goal of this PR was issue #149.


renner:
I merged it, just added one more commit with some minor changes as cdd1ed93636e7b36062c704ec8eb4dea64a79534. Thanks for this contribution, @admd!

admd:
Thank you @renner .

PR #199, opened 2017-04-30
fix sending kwarg in payload in RunnerCall

mcalmer:
using kwarg's' as keyword seems to be wrong and is not working anymore with salt 2016.11.4 .
It works with older versions but maybe only by accident. Using 'kwarg' seems to work with old and new versions.

Maybe somebody can check this with salt. The docs are not that clear.

### DIFF
--- a/src/main/java/com/suse/salt/netapi/calls/RunnerCall.java
+++ b/src/main/java/com/suse/salt/netapi/calls/RunnerCall.java
@@ -44,7 +44,7 @@ public class RunnerCall<R> implements Call<R> {
     public Map<String, Object> getPayload() {
         HashMap<String, Object> payload = new HashMap<>();
         payload.put("fun", functionName);
-        kwargs.ifPresent(kwargs -> payload.put("kwargs", kwargs));
+        kwargs.ifPresent(kwargs -> payload.put("kwarg", kwargs));
         return payload;
     }
 
--- a/src/test/resources/async_via_event_list_job_request.json
+++ b/src/test/resources/async_via_event_list_job_request.json
@@ -1,6 +1,6 @@
 [
   {
-    "kwargs": {
+    "kwarg": {
       "jid": "20161115135014396739"
     },
     "client": "runner_async",


mcalmer:
@dincamihai do you know if kwargs or kwarg is correct for RunnerCall?

admd:
Acording to @lucidd , in older version kwargs(Wheel & Runner) and kwarg(LocalCall) were not compatible. We couldn't find the reason why SALT had two different names.

In #193, I made some changes where during refactoring, i renamed kwarg to kwargs( should have been otherway around :) ). But on real salt instance,  started to have problem in localCall so we had to revert the related changes back. 
 
@mcalmer  Can you please also test WheelCall on salt 2016.11.4 because it also uses kwargs.  It would be good if all *Call use the same name because besides being less confusing, it will also give us an opportunity to extract some more code in a parent class.

lucidd:
@admd its even a bit more complicated then that. The other day i was looking again at the differences and the way it was so far is `LocalCall` has `kwarg`, `RunnerCall` has `kwargs` (at least used to have) and `WheelCall` has neither it directly embeds its keyword arguments into the general call next to `fun` and authentication etc. All of this was implemented 2015 so quite a few salt versions ago. If `RunnerCall` now supports `kwarg` that would be great to have it unified a bit more.

mcalmer:
@lucidd Well, also in older versions RunnerCall works with "kwarg" only. It seems both was working but kwarg seems to what should work. I tried with 2016.11.3 and there both was working fine.

lucidd:
@mcalmer it did not work back with 2015.x.x salt and we never tested it again afterwards up until now so somewhere in between they must have changed it to support both. Its nice that its now closer to `LocalCall` i wonder if anything else has changed related to other differences in argument passing.

whiteinge:
`kwargs` plural was initially a mistaken addition and was then put on a deprecation path IIRC. `kwarg` singular is the correct path forward as part of an effort to make RunnerClient have a consistent interface with LocalClient.

whiteinge:
WheelClient is still an outlier with having those values at the top-level and does not currently support `kwarg` (or `kwargs`).

lucidd:
@whiteinge awesome! thanks for the confirmation :smile: Is there maybe some issue open for the `WheelClient` we can reference to get notified when it changes?

whiteinge:
There isn't an issue that addresses this individually although it's mentioned in the mega-issue https://github.com/saltstack/salt/issues/32778 that is tracking the various inconsistencies.

Issue #190, opened 2017-02-15
Errors in Runner calls are not reported fully


moio:
Runner calls might return Python errors and stack traces as the result, eg.

```json
salt/run/20170212133049657941/ret       {
    "_stamp": "2017-02-12T11:30:51.183745", 
    "fun": "runner.jobs.list_jobs", 
    "jid": "20170212133049657941", 
    "return": "Exception occurred in runner jobs.list_jobs: Traceback (most recent call last):\n  File \"/usr/lib/python2.7/site-packages/salt/client/mixins.py\", line 348, in low\n    data['return'] = self.functions[fun](*args, **kwargs)\nFile \"/usr/lib/python2.7/site-packages/salt/runners/jobs.py\", line 333, in list_jobs\n
ret = mminion.returners['{0}.get_jids'.format(returner)]()\n  File \"/usr/lib/python2.7/site-packages/salt/returners/local_cache.py\", line 343, in get_jids\nfor jid, job, _, _ in _walk_through(_job_dir()):\n  File
\"/usr/lib/python2.7/site-packages/salt/returners/local_cache.py\", line 69, in_walk_through\n    for final in os.listdir(t_path):\nOSError: [Errno 2] No such file or directory: '/var/cache/salt/master/jobs/32'\n", 
    "success": false, 
    "user": "admin"
}
```

We get a gson conversion `Exception` in the assumption the result is expected to be an `Object` and not a `String`:

```
Stack trace:com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected
BEGIN_OBJECT but was STRING at line 1 column 14 path $.return[0]
```

It would be great if we had a meaningful `Exception` instead, especially one containing the Python stack trace that could be put in logs for later troubleshooting.

admd:
Hi @moio  , problem should have been resolved after  #200 . You can test it yourself and if it fulfills your requirement then  you can close this issue.

renner:
I think we can close it already as @mateiw has tested it successfully, thank you @admd for this patch! If the solution is not sufficient yet feel free to reopen the issue @moio.

moio:
It's good, thank you very much @admd!!!

PR #200, opened 2017-05-02
Better error handling in Runner and Wheel calls

admd:
In case of exception where parser is unable to translate response into a proper type, show the original response from Salt.

Related issues #190

### DIFF
--- a/src/main/java/com/suse/salt/netapi/calls/RunnerCall.java
+++ b/src/main/java/com/suse/salt/netapi/calls/RunnerCall.java
@@ -5,8 +5,8 @@ import static com.suse.salt.netapi.utils.ClientUtils.parameterizedType;
 import com.suse.salt.netapi.AuthModule;
 import com.suse.salt.netapi.client.SaltClient;
 import com.suse.salt.netapi.exception.SaltException;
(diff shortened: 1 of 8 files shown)


admd:
@lucidd  I might be missing something but would it make sense if return type of function e.g callSync in RunnelCall be like the Result<R> instead of R,  just like we have in LocalCall so we can get the advantage of 
Xor type.  I know Result<> is meant for minion result but can't we use it for Runner Calls as well.

lucidd:
@admd i think that makes total sense to have the same error handling in Runner and Wheel. I'm not sure tho to what extend the same error handling will work.

admd:
Thank you for your quick response.

admd:
Hi @lucidd , for synchronous methods, I have already made required changes for better error handling. But as asynchronous methods both in Wheel and Runner calls return scheduled job information immidately that we can later use to query the result, I can't think of much advantage here to return this result  wrapped in Result class.  

If you think  it is enough, can you please review my changes and if something is missing, please do let know.

admd:
Thank you.👍

mateiw:
Tested this and seems to work fine. 👍  from me too.

PR #203, opened 2017-06-05
Getter for function name.

admd:
A getter has been added to get just the function name by extracting it from function string( containing module and function name).  

Now there are two getters, one is getFunction() which returns the string composed of module and function name in form of module.function and other one is getFunctionName() which will return just the function name. 

Related issue #149

### DIFF
--- a/src/main/java/com/suse/salt/netapi/utils/ClientUtils.java
+++ b/src/main/java/com/suse/salt/netapi/utils/ClientUtils.java
@@ -69,34 +69,17 @@ public class ClientUtils {
     }
 
     /**
-     * Get the module name from a given function. In case of e.g. "test.ping", this method
-     * will return "test".
+     * Extract the module and function name from the function string based on '.' .
+     * In case of e.g. "test.ping", this method will return String array {'test','ping'}
      *
      * @param function string containing module and function name (e.g. "test.ping")
-     * @return the module name
+     * @return String array containing module name as 1st element and function name as 2nd
      * @throws IllegalArgumentException if a given function string does not contain a '.'
      */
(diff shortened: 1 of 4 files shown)


admd:
Hi @renner , thank you for the review.  Yes, it make perfect sense to move extraction of function name to constructor just like it is for moduleName.

src/main/java/com/suse/salt/netapi/utils/ClientUtils.java
```diff
@@ -83,4 +83,20 @@ public static String getModuleNameFromFunction(final String function)
...
+     * @return the function name
+     * @throws IllegalArgumentException if a given function string does not contain a '.'
+     */
+    public static String getFunctionNameFromFunction(final String function)
+            throws IllegalArgumentException {
+        if (!function.contains(".")) {
+            throw new IllegalArgumentException(function);
+        }
+        return function.split("\\.")[1];
+    }
```
renner: I would like to suggest one more change to the patch: In the current version we have two helper methods that are mostly equal to each other, just one of them returns the first and the other returns the second array element. In order to simplify this a bit, couldn't we just have one method `splitFunction(final String function)` that would return the array as returned by `function.split("\\.")`, and then in the constructor assign module and function name like this?

```java
String[] splitFunction = splitFunction(function);
this.moduleName = splitFunction[0];
this.functionName = splitFunction[1];
```

admd: Hi @renner , thank you again for the feedback. I have made the necessary changes.

renner:
I merged it, thanks for fixing it up!

Issue #165, opened 2016-07-07
Getting empty responses from a localCall.callSync


nw0428:
When I run the following code: `result = localCall.callSync(client, globTarget, saltUser, saltPassword, AuthModule.PAM);` result is an empty map.

renner:
Hey @nw0428, we need more information to further debug your issue: what exact type of a call is the `localCall` in your example? Could it possibly be the case that your `globTarget` does not actually match any minions? Does `PAM` authentication work for you otherwise? Please give enough code so that we can actually try to reproduce the issue, thanks!

osbominix:
The problem is reproducible if the globTarget does not match any minions. 

Like the example file [Calls.java](https://github.com/SUSE/salt-netapi-client/blob/master/src/test/java/com/suse/salt/netapi/examples/Calls.java) from package com.suse.salt.netapi.examples, 
if both minions("minion1", "minion2") are not running, the value of "grainResults" will be empty.

In this case, it could be helpful to show a warning message. For example, do the validation for the return value of callSync method in class LocalCall?

renner:
Closing the issue as this is actually the expected behavior for minions that are either down or minions that are not matching an existing target as clarified in [this comment](4190cde15fca64de79610ade31e92366038b16b8). Feel free to revive it in case you have further questions.

PR #204, opened 2017-06-22
Throw exception if response is empty by invoke method callSync

osbominix:
related issue: 
[Getting empty responses from a localCall.callSync #165](https://github.com/SUSE/salt-netapi-client/issues/165)

### DIFF
--- a/src/main/java/com/suse/salt/netapi/calls/LocalCall.java
+++ b/src/main/java/com/suse/salt/netapi/calls/LocalCall.java
@@ -374,13 +374,8 @@ public class LocalCall<R> extends AbstractCall<R> {
             final SaltClient client, Target<?> target,
             String username, String password, AuthModule authModule)
             throws SaltException {
-        Map<String, Result<R>> result = callSyncHelper(client, target, Optional.of(username),
-        Optional.of(password), Optional.of(authModule), Optional.empty()).get(0);
-        if (!result.isEmpty()) {
-                return result;
-        } else {
-                throw new SaltException("The target you defined does not match any minions.");
-        }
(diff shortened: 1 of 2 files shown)


lucidd:
whats the advantage of throwing an exception here? Having an empty result for a minion can mean multiple things. There is no minion that matches, the minion is down, the minion is unreachable, the call took to long and timed out. The salt api does not give us any information about what happened so we can not construct a single error message that fits all.

osbominix:
It is just because the usability. If a call has been invoked, there should be something in the response result. An empty result means multiple things, but it also means nothing. As a user who invokes the call and gets a empty result, it would be helpful for him to show some hints(as exception or log or something else) anywhere, like "Empty result, please check if there is no minion that matches, the minion is down, the minion is unreachable, or the call took to long and timed out. " It seems all callSync methods need a extension to handle this issue.  whatever, I am a newbie in this area of saltstack, maybe an empty result is not that bad in this case, if user doesn't care what is happened behind.

lucidd:
I think this handles a very narrow case. The only time this exception would be thrown is if the result is completely empty. In all other cases where only certain minions have no result but others have it would not be thrown. Errors happen on a minion by minion basis so they have to be checked and handled that way. In addition having a result does not mean you have no error you still need to check the result itself.
I agree that its a bit unintuitive for a user and it would probably be nice if we add some documentation about what a missing result means in terms of salt.

osbominix:
good, so the exception handling has been reverted. 

For the example file Calls.java which may return empty value, a comment has been added there.

renner:
Thanks for your efforts @osbominix, I merged your hint by squashing all your commits!

PR #196, opened 2017-03-31
Cmd module

smatei:
Fixed cmd.exec_code_all (should return Map<String, Object>).
Added cmd.exec_code, cmd.has_exec.
Added tests for cmd functions.

PR #209, opened 2017-07-12
Cmd module enhancements

renner:
This is a newer version of @smatei's #196: Commits were squashed together after rebasing on the current master and results are parsed into the already existing `CmdExecCodeAllResult` class instead of just a generic `Map<String, Object>` with 0d8ca8c.


renner:
Merged, thanks to @smatei for this contribution and @lucidd for the review!

PR #210, opened 2017-07-17
Allow use of XorTypeAdapter for general Xor types

moio:
Currently XorTypeAdapterFactory assumes the left side of the type is always a SaltError - that is currently the case but prevents using more complex Xor types, when returned data from Salt might come in different
'structures'.

This allows XorTypeAdapterFactory to work with any generic Xor type.


moio:
@lucidd swapped sides as per IRC chat. OK to merge?

PR #217, opened 2017-11-01
Allow custom lowstate data to be passed onto SaltClient.startCommand

### LINKED ISSUE #216: Allow setting custom lowstate data on startCommand
Currently, `SaltClient.startCommand` sets fixed lowstate data sent towards API, which limits the usage in certain way.

Following usecase specifies the needs for this.
As Job cache method [prep_jid](https://github.com/saltstack/salt/blob/develop/salt/returners/local_cache.py#L80) can accept `passed_jid` via lowstate data, users / systems might be able to craft custom JID before actual job execution, thus bypassing several race conditions when JID wasn't known prior to actual execution.

DarkKnightCZ:
Fixes #216

Test was created to verify the custom properties are present in request data. Since there is no need to verify the actual response (already handled in `testStartCommand`), I didn't include those asserts.

### DIFF
--- a/src/main/java/com/suse/salt/netapi/client/SaltClient.java
+++ b/src/main/java/com/suse/salt/netapi/client/SaltClient.java
@@ -278,7 +278,28 @@ public class SaltClient {
      */
     public <T> ScheduledJob startCommand(final Target<T> target, final String function,
             List<Object> args, Map<String, Object> kwargs) throws SaltException {
-        Map<String, Object> props = new LinkedHashMap<>();
+        return this.startCommand(target, function, args, kwargs, new HashMap<>());
+    }
+
+    /**
+     * Generic interface to start any execution command and immediately return an object
+     * representing the scheduled job.
+     * <p>
+     * {@code POST /minions}
+     *
+     * @param <T> type of the tgt property for this command
+     * @param target the target
+     * @param function the function to execute
+     * @param args list of non-keyword arguments
+     * @param kwargs map containing keyword arguments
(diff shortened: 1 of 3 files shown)


renner:
Hey @DarkKnightCZ, thanks a lot for contributing! While your patch is handling a valid use-case I would like to make you aware that we are actually planning to deprecate and remove the `SaltClient.startCommand()` method soon with the merge of #170. Background is that most of the endpoints in the Salt API (e.g. `/minions`) are just shortcuts to other Salt calls that can also be performed using the meanwhile unified API of the `LocalCall`, `WheelCall` or `RunnerCall` classes.

I would recommend you to see some examples [of how to use `LocalCall`](https://github.com/SUSE/salt-netapi-client/blob/master/src/test/java/com/suse/salt/netapi/examples/Calls.java) or [`WheelCall` objects](https://github.com/SUSE/salt-netapi-client/blob/master/src/test/java/com/suse/salt/netapi/examples/Wheel.java) and maybe try to implement your code in that way. Note that any Salt call can also be executed asynchronously (e.g. `Test.ping().callAsync()`) which should be the same thing that the existing `SaltClient.startCommand()` method would do.

We would be happy then to merge a patch to allow adding custom lowstate data for those classes (if not supported yet). Let us know what you think about this or ask me anything if you have any further questions.

renner:
Hey, I am closing this pull request now, sorry! We meanwhile merged #170 so your patch doesn't apply to the current master branch anymore as mentioned in my comment above. Feel free to redo your contribution following my recommendations above. Thank you!

Issue #103, opened 2015-09-02
Implement all expr_form types


lucidd:
[all expr_forms](https://docs.saltstack.com/en/latest/ref/clients/)
maybe some relevant [documenatation](https://docs.saltstack.com/en/latest/topics/targeting/compound.html)
- [ ] pcre - Perl style regular expression
- [x] grain - Match based on a grain comparison
- [ ] grain_pcre - Grain comparison with a regex
- [ ] pillar - Pillar data comparison
- [ ] pillar_pcre - Pillar data comparison with a regex
- [ ] range - Use a Range server for matching
- [ ] compound - Pass a compound match string

Work in progress  [branch](https://github.com/SUSE/saltstack-netapi-client-java/tree/full-expr-form-support)

mchugh19:
The code has changed lots since this was opened. Is the recorded state still accurate?

renner:
Unfortunately it looks like the checklist of unimplemented matchers is still accurate, even though a lot of other code has changed. The referenced [branch] (https://github.com/SUSE/saltstack-netapi-client-java/tree/full-expr-form-support) does not seem to exist anymore, though. @lucidd, does your patch live on somewhere else meanwhile?

lucidd:
the branch only contained grains so far and that part got merged.

mcalmer:
Seems that `expr_form` is deprecated and must be replaced very soon:

```
DeprecationWarning: The target type should be passed using the 'tgt_type' argument instead of 'expr_form'. Support for using 'expr_form' will be removed in Salt Fluorine.
```

salt is currently at `Oxygen` release and `Fluorine` will be the next one.

renner:
Closing this issue as we have #226 merged meanwhile. Created a [new issue](https://github.com/SUSE/salt-netapi-client/issues/103#issuecomment-352685882) with the contents of the last comment.

PR #218, opened 2017-11-01
Implement CompoundMatcher target

DarkKnightCZ:
This PR brings the possibility to use compound matchers when targeting minions (as discussed in #103), but changes ResultInfo API a bit due to different data returned from API.

As discussed in the issue, I couldn't really think of how this should be tested.

### DIFF
--- a/src/main/java/com/suse/salt/netapi/datatypes/target/CompoundMatcher.java
+++ b/src/main/java/com/suse/salt/netapi/datatypes/target/CompoundMatcher.java
@@ -1,0 +1,37 @@
+package com.suse.salt.netapi.datatypes.target;
+
+/**
+ *
+ * Target for referencing compound matcher.
+ */
+public class CompoundMatcher implements Target<String> {
+
+    private final String matcher;
+
+    /**
+     * Constructor expecting a matcher as string.
+     *
+     * @param matcher
+     *            the compound matcher as string
+     */
+    public CompoundMatcher(String matcher) {
+        this.matcher = matcher;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getTarget() {
+        return matcher;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getType() {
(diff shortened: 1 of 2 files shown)


renner:
I guess we could merge this proposal to initially represent the compound matchers, at least until someone takes the time to implement something more type safe. Instead of just taking a string (most of the matchers are just strings in the end) we were thinking to eventually implement an API that actually allows to combine instances of the other `Target` implementations (matchers) with `and`, `or` and `not` operators, maybe like this simple example:

```java
CompoundMatcher.and(new Glob("web*"), new MinionList("minion1", "minion2"));
```

Let me know what you think about it, eventually you might want to consider extending your patch but be aware that certainly more effort would be needed for that.

src/main/java/com/suse/salt/netapi/results/ResultInfo.java
```diff
@@ -34,7 +34,7 @@
     private String user;
 
     @SerializedName("Target")
-    private String target;
+    private Object target;
```
renner: Why does it need to be `Object` here, isn't a compound matcher also a `String` in the end (`Target<String>`)?

src/main/java/com/suse/salt/netapi/datatypes/target/CompoundMatcher.java
```diff
@@ -0,0 +1,37 @@
+package com.suse.salt.netapi.datatypes.target;
+
+/**
+ *
+ * Target for referencing compound matcher.
+ */
+public class CompoundMatcher implements Target<String> {
```
renner: Can compound matchers be used with Salt SSH? In this case you might want to also implement `SSHTarget<String>` here as for example in [`Glob`](https://github.com/SUSE/salt-netapi-client/blob/master/src/main/java/com/suse/salt/netapi/datatypes/target/Glob.java#L6).

renner:
Hey @DarkKnightCZ, I am sorry to close your PR but in the meantime we merged #226 which also adds support for the compound matchers.

PR #219, opened 2017-11-14
Support salt.runners.event + example code

renner:
This patch adds support for the [event runner module](https://docs.saltstack.com/en/latest/ref/runners/all/salt.runners.event.html) that allows to send custom events on the Salt event bus. There is also some example code showing how to call runner modules using the library.


renner:
Thanks @lucidd, merged.

PR #220, opened 2017-11-17
[WIP] Async and non-blocking HTTP backend

renner:
This patch is about replacing the existing HTTP client backend (that can run on either the Apache HTTP client or the JDK classes) with one that works in a non-blocking fashion, in this case the Apache [HttpAsyncClient](https://hc.apache.org/httpcomponents-asyncclient-dev/). The outcome of this is that we can actually benefit from the concurrency features of Java 8, especially [CompletableFuture](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html). All the main library methods will then return a [CompletionStage](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html) containing the result instead of just the result. This is why we are targeting the version 1.0.0 milestone with this patch.

### DIFF
--- a/src/main/java/com/suse/salt/netapi/client/SaltClient.java
+++ b/src/main/java/com/suse/salt/netapi/client/SaltClient.java
@@ -23,6 +23,13 @@ import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.reflect.TypeToken;
 
+import org.apache.http.HttpResponse;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.concurrent.FutureCallback;
+import org.apache.http.entity.ContentType;
+import org.apache.http.entity.StringEntity;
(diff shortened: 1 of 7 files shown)


src/main/java/com/suse/salt/netapi/client/SaltClient.java
```diff
@@ -56,42 +50,22 @@
...
-            ExecutorService executor) {
+    public SaltClient(URI url, ClientConfig config) {
+        this.config = config;
         // Put the URL in the config
         config.put(ClientConfig.URL, url);
-        this.connectionFactory = connectionFactory;
-        this.executor = executor;
+
+        // TODO: Replace connectionFactory with this
+        this.asyncConnectionFactory = new HttpAsyncClientConnectionFactory(config);
```
renner: Please remove the TODO comment here as the `connectionFactory` is gone now.

PR #222, opened 2017-11-22
Do not depend on javax.xml.bind any longer

renner:
For JDK9 `javax.xml.bind` is [deprecated to be removed](https://docs.oracle.com/javase/9/docs/api/java.xml.bind-summary.html) in a future version. Now as JDK8 finally includes Base64 encoding and decoding we should make use of that in order to avoid having to add an additional external library in the future.


juliogonzalez:
Don't we need to apply the same patch to other parts of the code?

This is what I found at the RPM package itself: http://pastebin.nue.suse.com/17430

juliogonzalez:
LGTM (previous comment was a mistake)

PR #225, opened 2017-11-29
Build with JDK9

renner:
This pull request is supposed to add support for building the library against JDK9 (in addition to JDK8).


renner:
Merged, thanks for your reviews!

PR #232, opened 2018-04-20
Change parameter name: 'expr_form' to 'tgt_type'

### LINKED ISSUE #227: Deprecation of the `expr_form` argument
The `expr_form` argument is deprecated and must be replaced very soon with `tgt_type`:

```
DeprecationWarning: The target type should be passed using the 'tgt_type' argument instead of 'expr_form'. Support for using 'expr_form' will be removed in Salt Fluorine.
```

Salt is currently at the `Oxygen` release and `Fluorine` will be the next one.

renner:
This patch is to fix #227: `expr_form` is deprecated and will be replaced with `tgt_type` everywhere.


mcalmer:
This is a bit dangerous. Does salt 2016.11.4 support it? If not, this lib requires minimal salt 2018? Or 2017.7? You should document it somewhere.

PR #234, opened 2018-05-18
Enable "Pkg.InfoInstalled" to return all installed package versions

meaksh:
This PR enables `Pkg.InfoInstalled` call to get all installed package versions on the minion (instead only the latest version) by introducing the `allVersions` parameter for `Pkg.InfoInstalled`.

Salt `all_versions` parameter for `pkg.info_installed` has been introduced here: https://github.com/saltstack/salt/pull/47678 (approved but not yet merged)

### DIFF
--- a/src/main/java/com/suse/salt/netapi/calls/modules/Pkg.java
+++ b/src/main/java/com/suse/salt/netapi/calls/modules/Pkg.java
@@ -264,15 +264,14 @@ public class Pkg {
      *
      * @param attributes list of attributes that should be included in the result
      * @param reportErrors if true will return an error message instead of corrupted text
-     * @param allVersions if true will return all installed package versions, not only latest
      * @param packages optional give package names, otherwise return info about all packages
      * @return the call
      */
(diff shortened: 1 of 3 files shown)


src/main/java/com/suse/salt/netapi/calls/modules/Pkg.java
```diff
@@ -259,6 +259,27 @@ private Pkg() { }
...
+    /**
+     * Call 'pkg.info_installed' API.
+     *
+     * @param attributes list of attributes that should be included in the result
+     * @param reportErrors if true will return an error message instead of corrupted text
+     * @param allVersions if true will return all installed package versions, not only latest
+     * @param packages optional give package names, otherwise return info about all packages
+     * @return the call
+     */
+    public static LocalCall<Map<String, Xor<Info, List<Info>>>> infoInstalled(List<String> attributes,
```
lucidd: like discussed i would remove the `allVersions` parameter here and pass always `true` since the return type only makes sense for `true` in case you would pass false you might as well use the old function with the simpler return type.

meaksh:
@lucidd changes done!

PR #236, opened 2018-07-03
remove client config and unify authentication methods

lucidd:
This PR removes the ClientConfig since its generally insufficient if you want to customize http client config. Instead a user can now configure the http client backend they want to use and then simply pass it to us. Since the token has since been in the configuration each call method now uses an AuthMethod parameter to decide how to authenticate the call. AuthMethod either contains the password authentication parameters or a token. Overall this PR leads to a lot less code more unification between call method variants and also gets rid of the non thread safety we had due to the mutable config.


src/test/java/com/suse/salt/netapi/examples/Async.java
```diff
@@ -15,13 +18,14 @@
     private static final String PASSWORD = "saltdev";
 
     public static void main(String[] args) {
+        CloseableHttpAsyncClient httpClient = TestUtils.defaultClient();
```
renner: If the `defaultClient()` is something that could be used in *production* I would suggest to move it away from `TestUtils` to a better place.

lucidd: i don't think its suitable for production since its disables all timeout which you probably don't want in a production environment.

lucidd:
@renner javadoc warning are addressed

src/test/java/com/suse/salt/netapi/examples/Async.java
renner: Can't or shouldn't we provide some reasonable defaults for those timeouts, for example the values that were used before? I think in order to make it easy for people to get started quickly we could provide reasonable defaults with this default client.

lucidd: i added a helper method for creating an async http client with the default timeout values used in our config before

src/main/java/com/suse/salt/netapi/client/impl/HttpAsyncClientImpl.java
```diff
@@ -18,59 +16,47 @@
...
-    private final String endpoint;
-
-    /** Configuration */
-    private final ClientConfig config;
-
-    /** Parser to parse the returned result */
-    private final JsonParser<T> parser;
+    public static HttpAsyncClient defaultClient() {
+        //TODO
+        return null;
```
renner: @lucidd: The default client can now be generated via `HttpClientUtils` as far as I understand, so should we remove the TODO and this whole method here?

src/test/java/com/suse/salt/netapi/examples/Async.java
renner: Nice, so should we use `HttpClientUtils.defaultClient()` here? Do we still need to have `TestUtils.defaultClient()`?

src/test/java/com/suse/salt/netapi/examples/Events.java
```diff
@@ -2,14 +2,21 @@
...
+import org.apache.http.client.config.CookieSpecs;
+import org.apache.http.client.config.RequestConfig;
+import org.apache.http.impl.nio.client.CloseableHttpAsyncClient;
+import org.apache.http.impl.nio.client.HttpAsyncClientBuilder;
+import org.apache.http.impl.nio.client.HttpAsyncClients;
 import com.suse.salt.netapi.event.WebSocketEventStream;
 
 import java.net.URI;
 
+
```
renner: Unnecessary new line here.

PR #238, opened 2018-08-21
EventStream refactoring

moio:
This is a set of 5 refactorings to detach `EventStream` from its WebSocket implementation and allow alternative implementations.

This is needed in order to replace reading from WebSockets completely in Uyuni where we plan to read them from a Postgres database filled by a custom Salt Engine.

Best reviewed commit-by-commit.

### DIFF
--- a/src/main/java/com/suse/salt/netapi/event/AbstractEventStream.java
+++ b/src/main/java/com/suse/salt/netapi/event/AbstractEventStream.java
@@ -67,7 +67,9 @@ public abstract class AbstractEventStream implements EventStream {
      */
     protected void notifyListeners(Event event) {
         synchronized (listeners) {
-            listeners.forEach(l -> l.notify(event));
+            for (EventListener listener : new ArrayList<>(listeners)) {
+                listener.notify(event);
+            }
         }
     }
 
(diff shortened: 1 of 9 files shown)


moio:
JFI: I pushed a rebased version of these changes against v0.14 in https://github.com/SUSE/salt-netapi-client/tree/v0.15.0-event-stream-refactoring

I am not opening a PR unless you tell me otherwise.

src/main/java/com/suse/salt/netapi/event/AbstractEventStream.java
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Removes all listeners.
+     *
+     * @param code an integer code to represent the reason for closing
+     * @param phrase a String representation of code
+     */
+    protected void clearListeners(int code, String phrase) {
+        synchronized (listeners) {
+            listeners.stream()
+                    .forEach(listener -> listener.eventStreamClosed(code, phrase));
```
renner: Just a nitpick: there is different indentation used here and above, maybe those expressions would even fit on one line? By the way: it seems that it is possible to omit the `stream()` and call `forEach()` directly on Java 8 `Iterable`s.

moio: Polish added

moio:
I added one extra commit to handle the case in which event listeners are removed from some method called by `notify`. We use this for the Salt Remote Command feature in Uyuni.

I am merging as soon as Travis is happy.

PR #239, opened 2018-09-05
AbstractEventStream: use copy-on-write list implementation

moio:
I verified in Uyuni that the previous implementation might lead to excessive monitor contention over the list, especially since iteration (in `notifyListeners`) requires a monitor.

In the application that I am looking at, we call `notifyListeners` from multiple threads, and this results in all threads but one being blocked while the notification is in progress.

This change uses a different synchronized Collection which has an extra cost on writes but never blocks reads.

### DIFF
--- a/src/main/java/com/suse/salt/netapi/event/AbstractEventStream.java
+++ b/src/main/java/com/suse/salt/netapi/event/AbstractEventStream.java
@@ -17,8 +17,8 @@ package com.suse.salt.netapi.event;
 
 import com.suse.salt.netapi.datatypes.Event;
 
-import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 /**
  * Implements basic listener bookkeeping for EventStream.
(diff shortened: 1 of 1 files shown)


src/main/java/com/suse/salt/netapi/event/AbstractEventStream.java
```diff
@@ -66,10 +61,8 @@ public int getListenerCount() {
      * @param event the event
      */
     protected void notifyListeners(Event event) {
-        synchronized (listeners) {
-            for (EventListener listener : new ArrayList<>(listeners)) {
-                listener.notify(event);
-            }
+        for (EventListener listener : new ArrayList<>(listeners)) {
```
lucidd: if i remember correctly the new `ArrayList` here was done in case the callback of a listener modifies the listener list resulting in a `ConcurentModificationException`. From the `CopyOnWriteArray` docs it seems that it does not have this problem. 

> ... and the iterator isguaranteed not to throw {@code ConcurrentModificationException}

moio: Right, fixing.

PR #247, opened 2019-01-10
support local async batching

lucidd:
adding the batch parameter to async calls as its gonna be supported in salt with https://github.com/saltstack/salt/pull/50546

### DIFF
--- a/src/main/java/com/suse/salt/netapi/calls/LocalCall.java
+++ b/src/main/java/com/suse/salt/netapi/calls/LocalCall.java
@@ -321,7 +321,24 @@ public class LocalCall<R> extends AbstractCall<R> {
      */
     public CompletionStage<List<Map<String, Result<R>>>> callSync(final SaltClient client, Target<?> target,
             AuthMethod auth, Batch batch) {
-        return callSyncHelperNonBlock(client, target, auth, Optional.of(batch));
+        return callSync(client, target, auth, Optional.of(batch));
+    }
+
+    /**
+     * Calls a execution module function on the given target with batching and
+     * synchronously waits for the result. Authentication is done with the token
+     * therefore you have to login prior to using this function.
+     *
+     * @param client SaltClient instance
+     * @param target the target for the function
+     * @param auth authentication credentials to use
+     * @param batch parameter for enabling and configuring batching
+     * @return A list of maps with each list representing each batch, and maps containing
+     * the results with the minion names as keys.
+     */
+    public CompletionStage<List<Map<String, Result<R>>>> callSync(final SaltClient client, Target<?> target,
+            AuthMethod auth, Optional<Batch> batch) {
+        return callSyncHelperNonBlock(client, target, auth, batch);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/main/java/com/suse/salt/netapi/datatypes/Batch.java
```diff
@@ -8,16 +12,43 @@
...
     private Batch(String batch) {
         this.batch = batch;
     }
 
+    private Batch(String batch, Double delay) {
+        this.batch = batch;
+        this.delay = Optional.of(delay);
+    }
+
+    public Batch delayed(Double seconds) {
```
renner: Here it might be really good to have some documentation about the `delayed` config parameter.

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
```diff
@@ -121,11 +120,20 @@ public LocalCall(String functionName, Optional<List<?>> arg,
      * @param client SaltClient instance
      * @param target the target for the function
      * @param auth authentication credentials to use
+     * @param batch parameter for enabling and configuring batching
      * @return information about the scheduled job
      */
-    public CompletionStage<LocalAsyncResult<R>> callAsync(final SaltClient client, Target<?> target, AuthMethod auth) {
+    public CompletionStage<LocalAsyncResult<R>> callAsync(final SaltClient client, Target<?> target, AuthMethod auth,
+            Optional<Batch> batch) {
```
chiaradiamarcelo: Nitpick: It would be nice to have an homogeneous interface for calling "callAsync" and "callSync" in batch mode. This requires an Optional<Batch>, while this https://github.com/SUSE/salt-netapi-client/blob/master/src/main/java/com/suse/salt/netapi/calls/LocalCall.java#L281 requires just a "Batch"

lucidd: ah thanks almost fotgot about this again. I will add it tomorrow. :+1:

lucidd:
@renner so far it looks from @chiaradiamarcelo tests that it is backwards compatible since salt just ignores the additional parameters.

chiaradiamarcelo:
> @renner so far it looks from @chiaradiamarcelo tests that it is backwards compatible since salt just ignores the additional parameters.

yes, I can confirm that

chiaradiamarcelo:
@lucidd shouldn't this PR include a class to map the 'Batch started' salt event ?

chiaradiamarcelo:
The following is an example of a 'batch-start' event json:

salt/batch/20190130132632115965/start	{
    "_stamp": "2019-01-30T12:26:32.178962", 
    "available_minions": [
        "mch-minion_1.tf.local"
    ], 
    "batch_delay": 1.0, 
    "client": "local_async", 
    "down_minions": [
        "mch-minion_2.tf.local"
    ],
    "eauth": "auto", 
    "metadata": {
        "suma-action-id": 3111
    }, 
    "password": "", 
    "username": "admin"
}

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
chiaradiamarcelo: done

src/main/java/com/suse/salt/netapi/calls/LocalCall.java
```diff
@@ -112,6 +111,7 @@ public LocalCall(String functionName, Optional<List<?>> arg,
         return payload;
     }
 
+
```
renner: This additional newline seems to be unneeded.

PR #248, opened 2019-01-21
[WiP] Pass kwargs in a Salt Flourine compatible way

amannch:
Flourine doesn't support passing arguments on a top level anymore.
This fix changes the way arguments are passed


amannch:
gitarro rerun continuous-integration/travis-ci/pr !!!

lucidd:
closing this for now since it does not look like salt actually fixed the way they pass arguments and it seems unlikely that they will right now.

PR #253, opened 2019-02-05
salt.modules.cmdmod additions

i-sergienko:
Added ```cmd.run_all```, ```cmd.script``` and ```cmd.script_retcode``` bindings to ```com.suse.salt.netapi.calls.modules.Cmd```. Added related tests.  
Moved common stub logic in ```com.suse.salt.netapi.calls.modules.CmdTest``` to a separate method (```mockOkResponseWith```).


src/main/java/com/suse/salt/netapi/calls/modules/Cmd.java
```diff
@@ -29,6 +31,13 @@ private Cmd() { }
                 new TypeToken<String>(){});
     }
 
+    public static LocalCall<CmdArtifacts> runAll(String cmd) {
+        LinkedHashMap<String, Object> args = new LinkedHashMap<>();
```
lucidd: i think we could replace this in all methods with `Collections.emptyMap`.

src/main/java/com/suse/salt/netapi/results/CmdArtifacts.java
```diff
@@ -1,9 +1,9 @@
 package com.suse.salt.netapi.results;
 
 /**
- * Result structure as returned by cmd.exec_code_all to be parsed from event data.
+ * Result structure as returned by cmd.exec_code_all, cmd.run_all, cmd.script to be parsed from event data.
  */
-public class CmdExecCodeAll {
+public class CmdArtifacts {
```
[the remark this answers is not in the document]
i-sergienko: Renamed to ```CmdResult``` as suggested.

src/main/java/com/suse/salt/netapi/calls/modules/Cmd.java
i-sergienko: Replaced empty map creation with ```Collections.emptyMap()```, and also replaced single pair map creations with ```Collections.singletonMap```.

PR #257, opened 2019-03-04
Batch: align parameters with implementation and naming

moio:
`Batch` parameters were not complete / their names were not aligned with current Salt implementation. This PR addresses those issues.

Compare with:

https://github.com/saltstack/salt/blob/09359322f652367244fccaa1ec0261bc50a29bc3/salt/cli/batch_async.py#L29-L33


moio:
I was missing an upstream PR, this is not actually needed. Closing.

PR #259, opened 2019-03-05
 Add pillar and data match modules

brunobottazzini:
- This patch also includes documentation for the methods inside
the Match module class.
- Add unit test case for the Match module

### DIFF
--- a/src/main/java/com/suse/salt/netapi/calls/modules/Match.java
+++ b/src/main/java/com/suse/salt/netapi/calls/modules/Match.java
@@ -17,8 +17,8 @@ public class Match {
 
     /**
      * Return True if the minion ID matches the given compound target
-     * @param tgt
-     * @param minionId
+     * @param tgt pillar target
+     * @param minionId optional id of the minion
      * @return the {@link LocalCall} object to make the call
      */
     public static LocalCall<Boolean> compound(String tgt, Optional<String> minionId) {
@@ -31,7 +31,7 @@ public class Match {
 
     /**
      * Return True if it matches the given compound target
-     * @param tgt
+     * @param tgt pillar target
      * @return the {@link LocalCall} object to make the call
      */
     public static LocalCall<Boolean> compound(String tgt) {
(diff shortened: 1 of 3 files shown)


src/test/java/com/suse/salt/netapi/calls/modules/MatchTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+    private static final int MOCK_HTTP_PORT = 8888;
+
+    private SaltClient client;
+
+    @Rule
+    public WireMockRule wireMockRule = new WireMockRule(MOCK_HTTP_PORT);
+
+    static final String JSON_CONFIG_MANAGE_MODE = ClientUtils.streamToString(
+            FileTest.class.getResourceAsStream("/modules/match/match_output.json"));
```
renner: Is there a reason why the name was chosen as `JSON_CONFIG_MANAGE_MODE `? I am only curious why it was not rather named `JSON_MATCH_OUTPUT` or similar.

src/main/java/com/suse/salt/netapi/calls/modules/Match.java
```diff
@@ -47,6 +76,39 @@
                 Optional.of(args), new TypeToken<Boolean>(){});
     }
 
+    /**
+     * Return True if the minion matches the given pillar target.
+     * The delimiter argument can be used to specify a different delimiter.
+     * @param tgt
+     * @param delimiter
```
renner: As you added javadoc comments I might suggest you to also add for completeness short descriptions of all the parameters as it is usually done, for example:

```java
@param tgt pillar target
@param delimiter optional different delimiter
```

brunobottazzini: I will add it.

src/test/java/com/suse/salt/netapi/calls/modules/MatchTest.java
brunobottazzini: Hello Renner, 

Sorry in the delay to answer your message. I was on Vacation :)...
You are right. I will change the name and update the PR. Today....

Thank you

src/main/java/com/suse/salt/netapi/calls/modules/Match.java
brunobottazzini: @renner, I just pushed everything... I hope it can me merged :)

renner:
Sorry for the long delay @brunobottazzini, finally merged it.

PR #261, opened 2019-03-11
Add test to timezone.get_offset API call

drago8888:
Signed-off-by: Drago8888 <rjmateus1985@gmail.com>


drago8888:
Challenge from Ricardo Mateus

src/test/java/com/suse/salt/netapi/calls/modules/TimezoneTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private static final int MOCK_HTTP_PORT = 8888;
+
+    private static final String JSON_NULL_RESPONSE = ClientUtils.streamToString(
+            TimezoneTest.class.getResourceAsStream(
+            "/modules/timezone/null_response.json"));
+
+    private static final String JSON_GETOFFESET_OK_RESPONSE = ClientUtils.streamToString(
+            TimezoneTest.class.getResourceAsStream(
+            "/modules/timezone/getOffset_ok.json"));
```
renner: I might suggest to rename that file in order to not mix underscore and camel case notations to `get_offset_ok.json`.

src/test/java/com/suse/salt/netapi/calls/modules/TimezoneTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+        LocalCall<String> call = Timezone.getOffset();
+        assertEquals("timezone.get_offset", call.getPayload().get("fun"));
+
+        Map<String, Result<String>> response = call.callSync(client,
+                new MinionList("minion1"), AUTH).toCompletableFuture().join();
+        assertEquals(JsonNull.INSTANCE,
+                ((JsonParsingError) response.get("minion1").error().get()).getJson());
+
+        // Test for successful responses
```
renner: I would suggest to test each case separately here, would you maybe want to move this second use case into a separate method?

src/test/java/com/suse/salt/netapi/calls/modules/TimezoneTest.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+public class TimezoneTest {
+
+    private static final int MOCK_HTTP_PORT = 8888;
+
+    private static final String JSON_NULL_RESPONSE = ClientUtils.streamToString(
+            TimezoneTest.class.getResourceAsStream(
+            "/modules/timezone/null_response.json"));
+
+    private static final String JSON_GETOFFESET_OK_RESPONSE = ClientUtils.streamToString(
```
renner: There is a typo in this name, should be `JSON_GETOFFSET_OK_RESPONSE`.

src/test/java/com/suse/salt/netapi/calls/modules/TimezoneTest.java
renner: May I suggest to rename that file in order to avoid mixing underscore with camel case notations to `get_offset_ok.json`?

drago8888: done. Thanks

src/test/java/com/suse/salt/netapi/calls/modules/TimezoneTest.java
drago8888: Done. Thanks

src/test/java/com/suse/salt/netapi/calls/modules/TimezoneTest.java
drago8888: I agree with you, I think each case should be in a separated method and that was my first approach. 
The reason behind my change of approach was to keep the consistency with the rest of the project test, with one test method per function with several scenarios.
However in this case following the right way instead of consistency seems not to bring any harm to the future of the project, therefore I will  split the test into two as suggested.

PR #262, opened 2019-03-11
add more functions to file module

drago8888:
Add support for the following functions:
- get_uid
- get_user
- is_link
- symlink
Signed-off-by: drago8888 <rjmateus1985@gmail.com>


drago8888:
Challenge from Ricardo Mateus

src/test/java/com/suse/salt/netapi/calls/modules/FileTest.java
```diff
@@ -392,4 +462,68 @@ public final void testRmdir() {
...
+                        .withBody(JSON_TRUE_RESPONSE)));
+
+        LocalCall<Boolean> call = File.symlink("/test/", "/link");
+        assertEquals("file.symlink", call.getPayload().get("fun"));
+
+        Map<String, Result<Boolean>> response = call.callSync(client,
+                new MinionList("minion1"), AUTH).toCompletableFuture().join();
+        assertTrue(response.get("minion1").result().get());
+
+        // false response
```
renner: For all these tests I would suggest to separate the tests for the `true` and `false` responses, and rather create one test method for each use-case. This ensures that all assertions are actually performed. Otherwise if the first assertion is failing, the second one would not even be executed. Also if this test for example would fail, you don't see directly which of the use-cases is broken (the `true` or the `false` response).

src/main/java/com/suse/salt/netapi/calls/modules/File.java
```diff
@@ -288,4 +318,30 @@ private File() { }
...
+     * @param src   Path to file or directory
+     * @param path  Path to symbolic link
+     * @return      The {@link LocalCall} object to make the call
+     */
+    public static LocalCall<Boolean> symlink(String src, String path) {
+        Map<String, Object> args = new LinkedHashMap<>();
+        args.put("src", src);
+        args.put("path", path);
+        return new LocalCall<>("file.symlink", Optional.empty(), Optional.of(args)
+                , new TypeToken<Boolean>(){});
```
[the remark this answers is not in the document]
drago8888: yeah. Done

src/test/java/com/suse/salt/netapi/calls/modules/FileTest.java
drago8888: I agree with you, I think each case should be in a separated method and that was my first approach. 
The reason behind my change of approach was to keep the consistency with the rest of the project test, with one test method per function with several scenarios.
However in this case following the right way instead of consistency seems not to bring any harm to the future of the project, therefore I will  split all the tests into suggested into one per scenario.

PR #268, opened 2019-04-10
Version 0.16.0

admd:
Update the checkstyle version and changes for the next release.


pom.xml
```diff
@@ -4,7 +4,7 @@
   <groupId>com.suse.salt</groupId>
   <artifactId>salt-netapi-client</artifactId>
   <packaging>jar</packaging>
-  <version>0.16.0-SNAPSHOT</version>
+  <version>0.16.1-SNAPSHOT</version>
```
juliogonzalez: This is releasing 0.16.1-snapshot, while title says 0.16.0

PR #270, opened 2019-04-10
Change version to 0.17.0-SNAPSHOT

admd:
Update the version to 0.16.1-SNAPSHOT in pom.xml


admd:
Thank you Johannes, I have updated it. BTW I couldn't release 0.16.0 in maven central because of credentials so when you feel better, can you please release 0.16.0 to the maven.

renner:
Sure, @admd, I am aware of that. Should be able to do it before the end of this week.

PR #281, opened 2020-02-27
Version 0.18.0-SNAPSHOT

admd:
Changes for the next release.


renner:
This patch now needs to be rebased on top of the current `master` branch that is now at the `v0.17.0` tag. Only the `pom.xml` file should actually be affected by this patch.

admd:
Thank you Johannes, I have rebased and pushed the branch.

PR #284, opened 2020-08-28
bring auth modules up to date with current salt docs

lucidd:
removes old modules that don't exist anymore and adds all currently documented ones.


moio (approved):
LGTM, thanks!

PR #287, opened 2020-09-22
Git module clone

paconte:
This patch add the git-clone functionality at the git module. The patch also fix an existing bug at the git module that throws a UnsopportedOperationException.

### DIFF
--- a/src/main/java/com/suse/salt/netapi/calls/modules/Git.java
+++ b/src/main/java/com/suse/salt/netapi/calls/modules/Git.java
@@ -4,7 +4,6 @@ import com.suse.salt.netapi.calls.LocalCall;
 import com.suse.salt.netapi.results.GitResult;
 import com.google.gson.reflect.TypeToken;
 
-import java.io.*;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -20,7 +19,7 @@ public class Git {
     private Git() { }
 
     public static LocalCall<GitResult> status(String cwd, Optional<String> user) {
-        List<String> args = Arrays.asList(cwd);
+        List<String> args = new ArrayList<>(Arrays.asList(cwd));
         user.ifPresent(usr -> args.add(usr));
         return new LocalCall<>("git.status", Optional.of(args),
                 Optional.of(emptyMap()), new TypeToken<GitResult>(){});
@@ -28,7 +27,7 @@ public class Git {
 
     public static LocalCall<String> add(String cwd, String filename,
             String opts, String gitOpts, Optional<String> user) {
-        List<String> args = Arrays.asList(cwd, filename, opts, gitOpts);
+        List<String> args = new ArrayList<>(Arrays.asList(cwd, filename, opts, gitOpts));
         user.ifPresent(usr -> args.add(usr));
         return new LocalCall<>("git.add", Optional.of(args),
                 Optional.of(emptyMap()), new TypeToken<String>(){});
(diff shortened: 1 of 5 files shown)


renner:
Thanks for this patch, @paconte! I will give it a review soon and hope to have it merged for our next release.

src/main/java/com/suse/salt/netapi/calls/modules/Git.java
```diff
@@ -43,9 +44,32 @@ private Git() { }
 
     public static LocalCall<Boolean> branch(String cwd, String branch,
             String opts, String gitOpts, Optional<String> user) {
-        List<String> args = Arrays.asList(cwd, branch, opts, gitOpts);
+        List<String> args = new ArrayList<>(Arrays.asList(cwd, branch, opts, gitOpts));
```
renner, line 47: I wonder why this would be needed (here and above), wouldn't `Arrays.asList()` already create and return a list of strings as `List<String>`?

renner (changes requested):
Looks good to me in general, just one question added and maybe indentation in the example code could be adjusted before merge (I wonder why checkstyle didn't find that)?

src/main/java/com/suse/salt/netapi/calls/modules/Git.java
paconte: Yes, you are right. It returns a list of strings as `List<String>`.  However the size of the list is fixed, this will not allow us to add other arguments to the list. Below the documentation of the method:
```

/**
     * Returns a fixed-size list backed by the specified array.  (Changes to
     * the returned list "write through" to the array.)  This method acts
     * as bridge between array-based and collection-based APIs, in
     * combination with {@link Collection#toArray}.  The returned list is
     * serializable and implements {@link RandomAccess}.
     *
     * <p>This method also provides a convenient way to create a fixed-size
     * list initialized to contain several elements:
     * <pre>
     *     List&lt;String&gt; stooges = Arrays.asList("Larry", "Moe", "Curly");
     * </pre>
     *
     * @param <T> the class of the objects in the array
     * @param a the array by which the list will be backed
     * @return a list view of the specified array
     */
    @SafeVarargs
    @SuppressWarnings("varargs")
    public static <T> List<T> asList(T... a) {
        return new ArrayList<>(a);
    }
```

If you add an element to that list, the application will throw an Exception. The current code actually send an exception, maybe it was not well tested. This is not the only line of code where this happens, so I made that change on in different places of the git module.

src/test/java/com/suse/salt/netapi/examples/GitModule.java
```diff
@@ -0,0 +1,60 @@
...
+    private static final String MINION_ID = "my_minion_id";
+    private static final String GIT_URL = "https://github.com/SUSE/salt-netapi-client.git";
+    private static final String CWD = "/home/saltuser/dev/";
+    private static final String NAME = "java_netapi";
+    private static final String HTTPS_USER = "saltgit";
+    private static final String HTTPS_PASS = "saltgit";
+
+    public static void main(String[] args) {
+		// Init the client
+    	SaltClient client =
```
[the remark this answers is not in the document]
paconte: Done: 4bf414e67a6620d6877eb85ac822bd894d59ba7d

src/main/java/com/suse/salt/netapi/calls/modules/Git.java
renner: Thanks, it makes sense!

renner (approved):

renner:
Merged for the [upcoming 0.18 release](https://github.com/SUSE/salt-netapi-client/milestone/14). Thanks again, @paconte!

PR #291, opened 2021-02-18
Feat: add pkg.list_patterns call

mbologna:
This PR adds support for calling `pkg.list_patterns`: https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.zypperpkg.html#module-salt.modules.zypperpkg

The `list_patterns` is only implemented by the `zypperpkg` "child" execution module of `pkg`: this is why the call returns an `Optional` map.

JUnit tests added.

### DIFF
--- a/src/main/java/com/suse/salt/netapi/calls/modules/Pkg.java
+++ b/src/main/java/com/suse/salt/netapi/calls/modules/Pkg.java
@@ -204,6 +204,28 @@ public class Pkg {
         }
     }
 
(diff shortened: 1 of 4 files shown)


lucidd (approved):

renner (approved):
It looks good to me :+1: Maybe the `PatternInfo` class could be moved to the results package (com.suse.salt.netapi.results`), would that make sense?

PR #292, opened 2021-05-18
add saltssherror for better error reporting

lucidd:
This adds a generic salt ssh error that will forward retcode and stderr returned in the salt ssh result in case the retcode is not 0.

### DIFF
--- a/src/main/java/com/suse/salt/netapi/errors/FunctionNotAvailable.java
+++ b/src/main/java/com/suse/salt/netapi/errors/FunctionNotAvailable.java
@@ -20,7 +20,8 @@ final public class FunctionNotAvailable implements SaltError {
     public <T> T fold(Function<FunctionNotAvailable, ? extends T> fnNotAvail,
             Function<ModuleNotSupported, ? extends T> modNotSupported,
             Function<JsonParsingError, ? extends T> jsonError,
-            Function<GenericError, ? extends T> generic
+            Function<GenericError, ? extends T> generic,
+            Function<SaltSSHError, ? extends T> saltSshError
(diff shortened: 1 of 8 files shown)


src/main/java/com/suse/salt/netapi/errors/FunctionNotAvailable.java
```diff
@@ -20,7 +20,8 @@ public String getFunctionName() {
     public <T> T fold(Function<FunctionNotAvailable, ? extends T> fnNotAvail,
             Function<ModuleNotSupported, ? extends T> modNotSupported,
             Function<JsonParsingError, ? extends T> jsonError,
-            Function<GenericError, ? extends T> generic
+            Function<GenericError, ? extends T> generic,
+            Function<SaltSSHError, ? extends T> saltSshError
```
renner, line 24: Very minor comment here: for more consistency as we use `SSH` in the class name (which I like), I would maybe use the same naming for the variable `saltSSHError`, starting with a lower-cased `s`.

src/main/java/com/suse/salt/netapi/errors/SaltSSHError.java
```diff
@@ -0,0 +1,40 @@
+package com.suse.salt.netapi.errors;
+
+import java.util.function.Function;
+
+/**
+ * Interface for all salt related errors
```
renner, line 6: The class comment should better be updated here.

renner (approved):
Looks really good to me as it is. Maybe some new unit tests could be useful for testing the `SaltSSHError` instances, if that is feasible.

PR #298, opened 2022-03-03
Use "stdout" as error message for SaltSSHError in cases where "stderr" is empty

meaksh:
This PR makes `SaltSSHError` to use `stdout` as error message in cases where the Salt SSH execution produces an error event with empty `stderr` but containing meaningful data in the `stdout`:

```json
{"uyuni-master-min-sles15.mgr.suse.de": {"stdout": "\rPassword: \nError: Wrong number of arguments!\n", "stderr": "", "retcode": 255, "id": "uyuni-master-min-sles15.mgr.suse.de", "fun": "state.apply", "jid": "20220303090427537110", "_stamp": "2022-03-03T09:04:30.263709"}}
```

We noticed this happening in some ocassions, specially when using "preflight script" on salt ssh, where `SaltSSHError` was raised without no message:

image

### DIFF
--- a/src/main/java/com/suse/salt/netapi/parser/ResultSSHResultTypeAdapterFactory.java
+++ b/src/main/java/com/suse/salt/netapi/parser/ResultSSHResultTypeAdapterFactory.java
@@ -85,7 +85,8 @@ public class ResultSSHResultTypeAdapterFactory implements TypeAdapterFactory {
                                         .flatMap(SaltErrorUtils::deriveError)
                                         .orElse(
                                                 new SaltSSHError(result.getRetcode(),
-                                                        result.getStderr().orElse(result.getStdout().orElse("")))
+							"stderr: \"" + result.getStderr().orElse("") +
+                                                        "\", stdout: \"" + result.getStdout().orElse("") + "\"")
                                         )
                         );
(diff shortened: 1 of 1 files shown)


src/main/java/com/suse/salt/netapi/parser/ResultSSHResultTypeAdapterFactory.java
```diff
@@ -85,7 +85,7 @@ public static boolean isResultSSHResult(Type type) {
                                         .flatMap(SaltErrorUtils::deriveError)
                                         .orElse(
                                                 new SaltSSHError(result.getRetcode(),
-                                                        result.getStderr().orElse(""))
```
vzhestkov, line 88: Is it possible that it return emptry string in case if "stderr" is present in the response, but blank?

vzhestkov: Agree and looking to the salt code most probably only one of them will contain something in case of error, and I like the idea to combine both to show :+1:

vzhestkov (approved):
:+1:

lucidd (approved):

PR #320, opened 2026-02-25
Fix test runtime errors when running on Java 17+

mackdk:
This PR replace the usages of `emptyMap()`, `singletonMap()` and `singletonList()` with the respective collection factory method. This avoids issue with gson serialization and private access on virtual machines that [enforce strong encapsulation](https://openjdk.org/jeps/403).


renner:
Shouldn't we also update the GH Actions workflow to use JDK 17 then, as well as `pom.xml` to update the `release` tag as part of the `maven-compiler-plugin`?

renner (approved):
It looks good to me.

PR #324, opened 2026-06-18
Migrate usage of internal GSON API

### LINKED ISSUE #323: Usage of internal GSON API prevents Dependency Upgrades
The internal GSON API used in this project was removed in version `2.9.0`. Using `com.google.gson.internal.*` prevents this project from upgrading to a newer GSON version. This is a blocker for us because the salt-netapi-client relies on such an old version of GSON while we use a newer version in our projec, resulting in dependency conflicts.

Affected classes:
- `CollectionTypeAdapterFactory`
- `OptionalTypeAdapterFactory`
- `ResultSSHResultTypeAdapterFactory`
- `XorTypeAdapterFactory`

SoulKa:
Closes #323

- migrate usages of the `com.google.gson.internal` packages to support the usage of newer GSON versions

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -128,7 +128,7 @@
     <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
-      <version>2.8.9</version>
+      <version>2.14.0</version>
     </dependency>
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -128,7 +128,7 @@
     <dependency>
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
-      <version>2.14.0</version>
+      <version>2.8.9</version>
     </dependency>
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
--- a/src/main/java/com/suse/salt/netapi/calls/LocalCall.java
+++ b/src/main/java/com/suse/salt/netapi/calls/LocalCall.java
@@ -174,9 +174,14 @@ public class LocalCall<R> extends AbstractCall<R> {
         Map<String, Object> customArgs = new HashMap<>();
         batch.ifPresent(v -> customArgs.putAll(v.getParams()));
 
+        Type asyncResultType = parameterizedType(null, LocalAsyncResult.class, getReturnType().getType());
+        Type listType = parameterizedType(null, List.class, asyncResultType);
+        Type wrapperType = parameterizedType(null, Return.class, listType);
+        @SuppressWarnings("unchecked")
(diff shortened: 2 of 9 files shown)


SoulKa:
Hi @renner, we currently struggle with using this project in our now-to-be spring boot 4 application. Is there still anything we need to do from our side in order for the GSON library to be upgrade here?

rjmateus:
@lucidd @mackdk do you think someone can have a look? To merge this PR we would then need to update the gson version on the depending project, right?

SoulKa:
@rjmateus if in doubt, I can remove the GSON version upgrade in this PR and only keep the migration of internal GSON calls. The salt API then works with GSON `2.8.9` and any newer version :)

rjmateus:
@SoulKa I think it would be better (and allow us to move faster) if the gson upgrade is done in a separate PR.

SoulKa:
@rjmateus thank you for your fast response in this matter. I reverted the GSON version upgrade and the tests still run locally.

mackdk (approved):
LGTM. 

The version bump is a kind of a blocker for us at the moment, since the version we use needs to remain in sync with what openSUSE provides. This change however works without any issue on `2.8.9` as well, and actually improves the existing code. Thanks for your contribution!

SoulKa:
The SaltClientDockerTest fails since `mbologna/*` images (from @mbologna) don't exist anymore and there seems to be no direct substitute. I'll try to find a saltstack testcontainer setup that works, so we have the tests running again if it blocks this PR.

mbologna:
> The SaltClientDockerTest fails since `mbologna/*` images (from @mbologna) don't exist anymore and there seems to be no direct substitute. I'll try to find a saltstack testcontainer setup that works, so we have the tests running again if it blocks this PR.

Apologies, by mistake I deleted mbologna/docker-salt. Now I revived and renovated it, and pushed the missing images to Docker Hub. All in all: you should relaunch the tests, the container pull will now work.
The tests should be adapted, though: 

https://github.com/SUSE/salt-netapi-client/blob/master/src/test/java/com/suse/salt/netapi/client/SaltClientDockerTest.java#L147

should now be `3006.0` as per the new Salt version in Leap

SoulKa:
Thank you very much @mbologna! I launched the containers locally and ran all tests. The `SaltClientDockerTest` work now, too.

SoulKa:
@mackdk are we ready to merge?

SoulKa:
@rjmateus anything left that blocks this PR? :)

rjmateus:
I think it's fine to merge. But we also need to prepare a version release.
@mackdk @lucidd do you think we can merge it?

SoulKa:
Hi @rjmateus and @mackdk, is there still a blocker for a (pre)release that includes this patch?

